PR #14, opened 2018-10-04
Changed everything to Recording and Sorting nomenclature

colehurwitz31:
* = Controversial and worth evaluating

Module: InputExtractor 
- InputExtractor --> RecordingExtractor
- *getEpochNames throws a not implemented error rather than returning an empty list
- *writeInput --> writeRecordedData
- getRawTraces --> getTraces
- getRawSnippets --> getSnippets

Module: SubInputExtractor
- SubInputExtractor --> SubRecordingExtractor

Module: MultiInputExtractor
- MultiInputExtractor --> MultiRecordingExtractor
- import spikeinterface as si --> from .RecordingExtractor import RecordingExtractor
- input_extractors --> recording_extractors

Module: OutputExtractor
- OutputExtractor  --> SortingExtractor
- *writeOutput --> writeSortedData

Module:SubOutputExtractor
- Made SubSortingExtractor a SortingExtractor object rather than an ABC object.
- *total_recording_frames added to SubOutputExtractor to replace need for RecordingExtractor. Throws value error if neither total_recording_frames or an end_frame is given.
- Current implementation appeared to be broken since the parent extractor was calling getUnitIds and also getNumFrames which would be neither a Recording or Sorting extractor.

Changed all examples and tests to reflect these changes. Currently, only two examples works as spikewidgets must be changed to reflect new changes and the biocam data requires access to a dataset the user won't have.

We also need unit tests for the Multi and subextractors since I am not confident they will all pass yet.

### DIFF
--- a/examples/demo_mearec_extractors.ipynb
+++ b/examples/demo_mearec_extractors.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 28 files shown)


alejoe91:
Hei @colehurwitz31 
Looks nice!

Why this:
*getEpochNames throws a not implemented error rather than returning an empty list
?

colehurwitz31:
>  *getEpochNames throws a not implemented error rather than returning an empty list

Previously, all getEpochsNames did was return an empty list no matter what. I think it was probably unfinished. I think it should be overwritten to return whatever variables the user may create for epochs.

PR #19, opened 2018-10-07
Added getUnitProperty to the sorting extractor.

mhhennig:
Purpose of this functionality is to retrieve information about sorted units, such as spike or unit locations.
The interface is based on dictionary keys, and information is stored in a dictionary within each class.

Added functionality to  read info to the HS2 sorting extractor, and to both write and read to the numpy class.

From the HS2 extractor, one can now obtain unit locations ``HS2e.getUnitProperty(u,'unit_location')`` and ``HS2e.getUnitProperty(u,'spike_locations')``

The numpy extractor also allows writing this info, e.g. ``NPe.setUnitProperty('a_new_property', data)``.

I'm neither sure about this implementation, nor about the naming, open for suggestions.

### DIFF
--- a/spikeinterface/SortingExtractor.py
+++ b/spikeinterface/SortingExtractor.py
@@ -54,6 +54,25 @@ class SortingExtractor(ABC):
         '''
         pass
 
+    def getUnitProperty(self, unit_id, property):
+        '''This function returns a property of a specified unit. It is a static
+        method so it can be used without instantiating this sorting extractor.
+
+        Parameters
+        ----------
+        unit_id: int
+            The id that specifies a unit in the recording.
+        property: string
+            The property requested from the extractor. Note individual extractors provide different properties.
+        Returns
+        ----------
+        data: numpy.ndarray
+            A numpy array containing the requested property.
+        '''
+        raise NotImplementedError("The getUnitProperty function is not \
+                                  implemented for this extractor")
+
+
     @staticmethod
     def writeSorting(self, sorting_extractor, save_path):
         '''This function writes out the spike sorted data file of a given sorting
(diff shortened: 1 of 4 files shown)


colehurwitz31:
So I started working on this issue and here is my implemenation:

```python
def __init__(self):
        self._unit_properties = {}

def addUnitProperty(self, property_name, property_data):
        '''This function adds a unit property data set under the given property
        name

        Parameters
        ----------
        property_name: str
            A property stored by the sorting extractor (pca_features, etc.)
        property_data
            The data associated with the given property name. Could be many
            formats as specified by the user.
        '''
        self._unit_properties[property_name] = property_data

    def getUnitProperty(self, property_name):
        '''This function rerturns the data stored under the property name given

        Parameters
        ----------
        property_name: str
            A property stored by the sorting extractor (pca_features, etc.)

        Returns
        ----------
        property_data
            The data associated with the given property name. Could be many
            formats as specified by the user.
        '''
        if(property_name in list(self._unit_properties.keys())):
            return self._unit_properties[property_name]
        else:
            raise ValueError("This property has not been added")
```

magland:
I think it looks good.

mhhennig:
Just fixed issues that led unit test to fail. Can this be merged? Cole, I think we wrote the same add/get-property functions, just slightly different documentation. Can you merge to get the best of the to?

colehurwitz31:
So I added the functionality of what you wrote into the parent class so that all you have to do is add information the way the parent class does. getUnitProperty and addUnitProperty have default implemenations now which is basically adding to a dictionary of dictionaries indexed by key and then by property_name

colehurwitz31:
So, I think I will push my branch and you can pull it and make your specific subclasses adhere to the default implemenation

Issue #35, opened 2018-10-10
Proposal: getSnippets -- replace snippet_len_before and snippet_len_after


magland:
snippet_len_before and snippet_len_after is cumbersome, I'd rather just have

snippet_len

which would be a single integer, with the option of passing in a 2-vector, which would represent the [len_before, len_after]

colehurwitz31:
So how would this be implemented in the arguments?

colehurwitz31:
Ok. That looks alright. I can do that on the Cole branch today and change the unit tests.

colehurwitz31:
I think snippet_offset can be added or subtracted from the reference frame before being passed into get snippets. So we should just keep reference frames and have the user pass in: spike_peaks - offset if they want.

mhhennig:
Yes, good point!

Issue #33, opened 2018-10-10
Proposal: setChannelProperty/getChannelProperty rather than getChannelInfo


magland:
In RecordingExtractor, some methods are overridable (expected to be overwritten in a subclass), and some are not. There are at least two reasons to make a method overwritable. But I think in the case of getChannelInfo, neither one of them apply. Therefore, to conform with the getUnitProperty/setUnitProperty method of SortingExtractor, I propose we replace getChannelInfo by

setChannelProperty(property_name, value)
getChannelProperty(property_name)

One special property will be 'location', which returns a 2-vector or 3-vector with the electrode coordinates.
Another property will be 'group', an integer specifying the group of the electrode.

magland:
Note that setChannelProperty will often be called in the constructor of the subclass, just like addEpoch is called in the MultiRecordingExtractor

colehurwitz31:
Same implementation and naming scheme as the unit property then?

alejoe91:
I agree too

PR #38, opened 2018-10-11
Cole

colehurwitz31:
I fixed two issues: #35 and #33 

I added unit tests for the channel property method and also changed the unit property arguments to match the new channel property methods (property_data --> value).

Ready to merge when approved.


magland:
Looks good to me. Merging.

PR #39, opened 2018-10-11
improve mda extractor and allow unit properties to bet get/set for multiple units at once

magland:
improve mdaextractor behavior for remote files
enable setting property of multiple units simultaneously
implement getUnitPropertyNames
allow getUnitProperty for multiple units

### DIFF
--- a/spikeinterface/SortingExtractor.py
+++ b/spikeinterface/SortingExtractor.py
@@ -62,20 +62,13 @@ class SortingExtractor(ABC):
         Parameters
         ----------
         unit_id: int
-            The unit id for which the property will be added
-            (or a list of ids if you want to simultaneously set the property
-            for multiple units)
+            The unit id for which the property will be set
         property_name: str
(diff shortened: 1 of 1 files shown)


colehurwitz31:
I don’t know about unit_id being a list also. It would have to be unit_ids i think

magland:
I think we should make an additional function instead -- I am beginning to use the framework and do think that getting/setting multiple at a time is important.

How about one of the following:

sorter.setUnitsProperty(unit_ids, property_name, values)
sorter.setUnitPropertyMulti(unit_ids, property_name, values)
(I don't think setUnitProperties is correct, although it sounds better)

I vote for setUnitsProperty

magland:
Yes. I'll adjust the pull request accordingly and then we can discuss.

magland:
I updated the pull request accordingly (although haven't yet implemented getChannelsProperty)

PR #41, opened 2018-10-13
Alessio

alejoe91:
Implemented `loadProbeFile` and `saveProbeFile` + some fixes.

- Probe files can be eirther `.prb` (location and group is saved - if available) or `.csv` (only location is saved)

- Added `getChannelPropertyNames()` in recording extractor and fixed a recursive issue in `getUnitPropertyNames()`

- some fixes to MeaRecExtractors that match new (and hopefully stable) implementation


colehurwitz31:
The code looks great! However, could you add more documentation to the probe file for what each function is doing specifically? For example, explain how csv and prb files are treated in the load function. The explanations are a bit sparse. Also, if you could remove the commented out code that would be better.

Otherwise, it looks good! Fyi, I will probably remove some of the variable names like ret and temp in the main code later.

PR #50, opened 2018-10-17
kilosort, klusta, spykingcircus sorting extractors + writeBinaryDatFormat function

alejoe91:
implemented extractors and tests for:

- klusta
- kilosort
- spykingcircus

+ in the tools the `writeBinaryDatFormat(recording, save_path, transpose=False, dtype='float32')` writes a dat file from a recording extractor


colehurwitz31:
Looks good! Do you think the binary writer should be encapsulated in a recording extractor called BinaryRecordingExtractror?

alejoe91:
I think that a function is enough, as the binary format has no info on sampling frequency for example,
so it would be useless on its own!

PR #53, opened 2018-10-21
Created getUnitSpikeFeatures and getUnitSpikeFeatureNames. Unimplemented

colehurwitz31:
Added:

getUnitSpikeFeatures(self, unit_id, feature_name, start_frame=None, end_frame=None):

getUnitSpikeFeatureNames(self, unit_id):

They both throw unimplemented errors if not implemented and their implemenation is up to the discretion of the subclasses.

### DIFF
--- a/spikeinterface/SortingExtractor.py
+++ b/spikeinterface/SortingExtractor.py
@@ -89,20 +89,16 @@ class SortingExtractor(ABC):
         raise NotImplementedError("The getUnitSpikeFeatures function is not \
                                   implemented for this extractor")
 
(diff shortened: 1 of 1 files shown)


alejoe91:
Looks good! I'll merge ;) Sure @magland would agree

magland:
Looks good. But I would propose a couple changes.

getUnitSpikeFeatureNames()
- default return an empty list
- do not take unit_id as input -- should return feature names across all the units.

colehurwitz31:
I agree. I will commit and push this tomorrow.

PR #57, opened 2018-10-27
Added getChannelIds and updated probe file handling

alejoe91:
- added `getChannelIds()`: if `self._channel_ids` is None, it returns `list(range(num_channels))`, otherwise it returns the list of channel_ids. Channels are handled as units in the SortingExtractors. The only way to change the channel ids is creating a `SubRecordingExtractor` with the `channel_ids` argument.

- updated `loadProbeFile()`:
    - returns a subrecording extractor
    - if 'channels' is in the probe file, it is considered the channel map
    - other properties can be loaded either by passing a dictionary channel_id: property, or a list with the length of the number of channels in the channel group
    - when loading locations from csv file, one can pass the `channel_map` and  `channel_group` arguments to load map and group information

- fixed some minor thing in Recording and Sorting extractor: `isinstance(id, (int, np.integer))` handles all possible numpy integers

### DIFF
--- a/spikeinterface/MultiRecordingExtractor.py
+++ b/spikeinterface/MultiRecordingExtractor.py
@@ -83,9 +83,6 @@ class MultiRecordingExtractor(RecordingExtractor):
     def getChannelIds(self):
         return list(range(self._num_channels))
 
-    def getNumChannels(self):
-        return self._num_channels
-
     def getNumFrames(self):
         return self._num_frames
 
--- a/spikeinterface/MultiRecordingExtractor.py
+++ b/spikeinterface/MultiRecordingExtractor.py
@@ -80,6 +80,9 @@ class MultiRecordingExtractor(RecordingExtractor):
         )
         return np.concatenate(list,axis=1)
 
+    def getChannelIds(self):
+        return list(range(self._num_channels))
+
     def getNumChannels(self):
         return self._num_channels
 
--- a/spikeinterface/RecordingExtractor.py
+++ b/spikeinterface/RecordingExtractor.py
@@ -50,17 +50,6 @@ class RecordingExtractor(ABC):
         pass
 
     @abstractmethod
-    def getNumChannels(self):
-        '''This function returns the number of channels in the recording.
-
-        Returns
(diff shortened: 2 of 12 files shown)


colehurwitz31:
Looks good! 

Couple of questions:

- Do you think that channel_map should be called channel_ids for consistency with the SubRecordingExtractor?

- I am not sure we should keep getNumChannels. Jeremy gave an example where it could be confusing in the SubRecordingExtractor where you iterate through range(rx.g
etNumChannels()) and it doesn't match the channel_ids.

magland:
Thanks Alessio and Cole. I'll comment on Monday (today I need to catch up
on some other things).

magland:
We should be consistent between getChannelIds() and getUnitIds(). The way you have it, @alejoe91, getChannelIds() is not implemented in the subclass, whereas getUnitIds() is implemented in the subclasses. Furthermore, IMO, SubRecordingExtractor should not access the private member _channel_ids of  the parent class. I think the right thing to do is to copy what we've already settled upon the the sorting extractor. Tag: @alejoe91 @colehurwitz31

colehurwitz31:
I think I agree with @magland that the best way to solve this now is just to leave getChannelIds as an abstract method that needs to be overwritten by the subclass and to get rid of the channel_ids internal variable 

Then the subclass can return whatever ids it wants and those can be reordered by a SubRecordingExtractor. 

It’s a bit more work in the subclasses, but it is consistent with the sorting extractor and should be just as easy to use. @alessio @magland. It also leaves the parent class a little cleaner. Hopefully, we can solve this issue and then SpikeInterface is basically done :)

alejoe91:
@colehurwitz31 @magland this should do it! If you agree I'll merge later :)

colehurwitz31:
Excellent!! Thanks for doing this Alessio. I know this one was a grind, but glad to be done :)

colehurwitz31:
Ready to merge @alejoe91? I think it looks all done.

alejoe91:
Yep, I'll merge! We cna do minor fixes later if needed

PR #58, opened 2018-10-29
Implemented UnitSpikeFeature methods and PhySortingExtractor

alejoe91:
Added implementation on sorting extractor of:
- `setUnitSpikeFeatures(self, unit_id, feature_name, value)`
- `getUnitSpikeFeatures(self, unit_id, feature_name, start_frame=None, end_frame=None)`
- `getUnitSpikeFeatureNames(self, unit_id=None)`

Implemented PhySortingExtractor to load files saved by `exportToPhy` from spiketoolkit


colehurwitz31:
Sorry, this was never merged. Was all of this added elsewhere? We already have the getUnitSpike stuff, but I am not sure about the exportToPhy and other stuff

alejoe91:
Hi @colehurwitz31. Thi was after the previous PR, so like yesterday :)
Actually the getUnitSpikeFeature stuff was basically not implemented in the base class, but I think that implementing it similarly to getUnitProperty makes sense, as the are going to be similar. You can see some use of the setUnitSpikeFeature in the PhySortingExtraxctor.

The exportToPhy is in spiketoolkit!

colehurwitz31:
I thought we agreed to leave the unit spike features to the subclasses because people may store features in a file and may not want to keep it in memory. 

Unit properties are usually pretty small and easy to store, but features for each spike are much trickier.

colehurwitz31:
Also, unit spike features can't really be set and the way you have implemented it doesn't allow for each spike to have a different feature unless I am misreading it. I think the current solution is best.

alejoe91:
@colehurwitz31 I think you are misreading it.. the feature should be a vector with the same length as the getUnitSpikeTrain. And you can load as many fetaures you want: in the phy sorting extractor I load pc_features and amps (https://github.com/colehurwitz31/spikeinterface/blob/alessio/spikeinterface/extractors/physortingextractor/physortingextractor.py#L39-L40).

I don't really see to problem of the files..it is up to the sorting extractors or to the user to add features or not, and I think that features are general for spike trains so they should be implemented in the base class. 

@magland ??

PR #59, opened 2018-11-08
Changed getTraces signature to match with the rest of SpikeInterface

colehurwitz31:
All other functions in SpikeInterface have channel_ids (or unit_id) as the first argument in the signature except the getTraces function. I changed all getTraces functions in the tests, extractors, and parent classes including the multi and sub extractors.

Original:
```
def getTraces(self, start_frame=None, end_frame=None, channel_ids=None)
```

Proposed change
```
def getTraces(self, channel_ids=None, start_frame=None, end_frame=None)
```

Other examples
```
def getUnitSpikeTrain(self, unit_id, start_frame=None, end_frame=None):

def setChannelProperty(self, channel_id, property_name, value):
```

### DIFF
--- a/spikeinterface/MultiRecordingExtractor.py
+++ b/spikeinterface/MultiRecordingExtractor.py
@@ -59,7 +59,7 @@ class MultiRecordingExtractor(RecordingExtractor):
         ind=inds[-1]
         return self._RXs[ind], ind, time-self._start_times[ind]
 
-    def getTraces(self, start_frame=None, end_frame=None, channel_ids=None):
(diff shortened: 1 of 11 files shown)


magland:
Looks good to me... I tend to use * in most of my functions, particularly those that could have some ambiguity in the ordering, so that it forces the call to explicitly name the arguments.

colehurwitz31:
I see, that's a good point. I don't it is necessary to have the * in this case though since we don't have it elsewhere in SpikeInterface. I am going to merge

PR #73, opened 2019-02-07
tridesclous extractor.

samuelgarcia:
tridesclous extractor.

### DIFF
--- a/spikeextractors/__init__.py
+++ b/spikeextractors/__init__.py
@@ -19,6 +19,7 @@ from .extractors.openephysextractors.openephysextractors import OpenEphysRecordi
 from .extractors.physortingextractor.physortingextractor import PhySortingExtractor
 from .extractors.bindatrecordingextractor import BinDatRecordingExtractor
 from .extractors.spykingcircussortingextractor.spykingcircussortingextractor import SpykingCircusSortingExtractor
+from .extractors.tridesclousextractor.tridesclousextractor import TridesclousSortingExtractor
 
 from . import example_datasets
 from .tools import loadProbeFile, saveProbeFile, writeBinaryDatFormat, getSubExtractorsByProperty
--- a/spikeextractors/extractors/tridesclousextractor/__init__.py
+++ b/spikeextractors/extractors/tridesclousextractor/__init__.py
@@ -1,0 +1,1 @@
+from .tridesclousextractor import TridesclousSortingExtractor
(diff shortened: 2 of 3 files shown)


colehurwitz31:
Hey Sam,

Your extractor looks great. Would it be possible to implement the writeSorting function for tridesclous also? It allows for converting between different sorted file formats.

Here is the documentation for the code:

```python
 @staticmethod
  def writeSorting(sorting, save_path):
      '''This function writes out the spike sorted data file of a given sorting
      extractor to the file format of this current sorting extractor. Allows
      for easy conversion between spike sorting file formats. It is a static
      method so it can be used without instantiating this sorting extractor.
      Parameters
      ----------
      sorting: SortingExtractor
          A SortingExtractor that can extract information from the sorted data
          file to be converted to the new format.
      save_path: string
          A path to where the converted sorted data will be saved, which may
          either be a file or a folder, depending on the format.
      '''
```

samuelgarcia:
I can try to make this **writeSorting**.
I don't anderstand the usecase : is it to inject the result of another sorters inside tyridesclous structure ?

There are several concern about this:
  * in tridesclous, the result is not only index+label for each spike but also the jitter, a flloat between -0.5 
    and +0.5. This represent the position of the true peak value in bewteen samples. This internally 
    mandatory 
  * If tridesclous have already launch the full spike sorting chain this is not difficult to replace the "spikes"
    vector. But if the "save_path" is fresh, then tridesclous need to run everything (filtering, pcas, ...), 
    because many internal data are also mandatory. Otehrwise the "working" would be corrupted
    So this function would need some parameters about everything. So this signature is not enough.

colehurwitz31:
Hi Sam,

The purpose of writeSorting is to take any abitrary sorting extractor (which may be extracting from any file format) and to write out the sorting result into the trideclous results format. This allows for conversions between different results file formats.

This is an optional function so it does not have to be implemented if it is not possible to do so. Maybe it could be possible to implement this if you calculated the jitter, ran pca, and did filtering within this function? Maybe you could also fill the different internal data with dummy values and then just have an explanation that the format doesn't have enough information to completely fill the trideclous format?

Otherwise, don't worry about implementing it, as it is not required (just nice to have).

samuelgarcia:
Hi Cole.
I won't integrate writeSorting right now it is too difficult in actual state.
So for me it is ready to merge.

Sam

colehurwitz31:
Great! Merging

samuelgarcia:
Thanks.

PR #90, opened 2019-02-28
Edit README, update mda extractors, and remove dependencies on kbucket and mountainlab_pytools

magland:
@colehurwitz31 and all, please take a look at this pull request.

I have revised some of the language in the readme. Feel free to reject or accept any of those modifications.

I have also updated the mda extractors, in preparation to using this repo as a dependency of spikeforest rather than having my own snapshot.

Importantly we will no longer have dependencies on mountainlab_pytools and kbucket.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.egg-info
 .ipynb_checkpoints
 __pycache__
+sample_*_dataset
--- a/README.md
+++ b/README.md
@@ -6,17 +6,17 @@ Version 0.3.0
 
 # spikeextractors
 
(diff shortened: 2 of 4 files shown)


colehurwitz31:
The README looks great :)

For the changes to the mda extractors, are there any dependencies besides numpy? Originally, there was a method to import the modules if the user had them installed so that all the dependencies depended on specific extractor being used. Is this still the case for mda?

magland:
I believe there are no dependencies aside from numpy, when working with local files.

For loading remote kbucket files there is a notification exception that explains what to install. But that module is not let available on pypi. But this is necessary for use within spikeforest.

colehurwitz31:
Great! I will merge

Issue #92, opened 2019-03-01
BinDatRecordingExtractor should have a "offset" or "byteoffset" keyword


samuelgarcia:
Some binary dat file do not start at 0.
For instance this [data](https://zenodo.org/record/1205233#.XHkWSHWLTIH) have an offset.

numpy.memmap offer a **offset** karg, BinDatRecordingExtractor offer possibility to be different from 0.

PR #93, opened 2019-03-01
Add offset to BinDatRecordingExtractor

samuelgarcia:
This implement #92.


colehurwitz31:
Great! Thanks Samuel :) nice catch

PR #94, opened 2019-03-01
Bug: wrong logic in loadProbeFile.

samuelgarcia:
It was testing the wring way.

### DIFF
--- a/spikeextractors/tools.py
+++ b/spikeextractors/tools.py
@@ -58,11 +58,11 @@ def loadProbeFile(recording, probe_file, channel_map=None, channel_groups=None):
                 for key_prop, prop_val in cgroup.items():
                     if key_prop == 'channels':
                         ordered_channels = np.concatenate((ordered_channels, prop_val))
-
+            
             if list(ordered_channels) == recording.getChannelIds():
                 subrecording = recording
             else:
-                assert np.all([chan in ordered_channels for chan in recording.getChannelIds()]), \
+                assert np.all([chan in recording.getChannelIds() for chan in ordered_channels]), \
                     "all channel_ids in the 'channels' section of the probe file " \
                     "must be in the original recording channel ids"
                 present_ordered_channels = [chan for chan in ordered_channels if chan in recording.getChannelIds()]
--- a/spikeextractors/tools.py
+++ b/spikeextractors/tools.py
@@ -62,9 +62,8 @@ def loadProbeFile(recording, probe_file, channel_map=None, channel_groups=None):
             if list(ordered_channels) == recording.getChannelIds():
                 subrecording = recording
             else:
-                assert np.all([chan in recording.getChannelIds() for chan in ordered_channels]), \
-                    "all channel_ids in the 'channels' section of the probe file " \
-                    "must be in the original recording channel ids"
+                if not np.all([chan in recording.getChannelIds() for chan in ordered_channels]):
+                    print('Some channel in PRB file are in original recording')
                 present_ordered_channels = [chan for chan in ordered_channels if chan in recording.getChannelIds()]
                 subrecording = SubRecordingExtractor(recording, channel_ids=present_ordered_channels)
             for cgroup_id in groups:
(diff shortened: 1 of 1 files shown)


samuelgarcia:
So there is a problem.

In most cases, I use (and same for pierre with circus), you have nb_channel=256 in the file.
But only nb_chan_in_group = 252. Because 4 channel are used for something else.

In this case all channel from the group belong to channel in files but 4 channel in file among do **not** belong to the group.

They way, we remove bad channel is not by channging the file (the recording) but by removing the chann in the PRB file. This is very faster (no memcopy).

In short example with a 8 channel in recording. 2 tetrodes. but one remove channel.
In recording :  [0, 1, 2, 3, 4, 5, 6, 7].
In prb:
  * group 0: [0, 2, 3]
  * group 1: [4, 5, 6, 7]

In this case the assertion is wrong.

**Also note**:  for the PRB channel are the absolut index of channel **inside** the file (= **inside** the recording). It is not an ID but the zero bassed absolut index.
So if you remove a channel from a recording, then the PRB is total wrong because global indexes have changed.

alejoe91:
not if you use the SubRecordingExtractor :)
See this example: https://github.com/CINPLA/expipe-plugin-cinpla/blob/dev/expipe_plugin_cinpla/scripts/openephys.py#L114-L122

The channel ids are preserved if you use the subrecordingextractor.

I think that a solution would be allow both behaviors. Not that the loadProbeFile is only used to load group and location information. With the current implementation, to actually remove channels one would need to use the SubRecordingExtractor. We can make it so that if not the probe_channels < recording_channel a SubRecordingExtractor is returned.

What do you think?

alejoe91:
After loading the probe_file,  a new probeFile will be written based on the information loaded. So in your case if the probe file has 252 channels and the recording has 256, then a sub recording extractor will be returned and the newly written probe file will have nb_channels = 252.

Alternatively we can do this: if a probe channel is not present in the recording channel the group is -1. Then only groups >= 0 are rewritten in the saveProbeFile. I'm ok with both implementation, but not with the first one as it would need to deal with missing channels in a different way.

samuelgarcia:
Yes channel ids are preserved but in PRB theses are are ID but index.

I am facing the problem of this, in mea_256.prb some channel are missing so the assertion is wrong from the beginig.
subrec = se.loadProbeFile(rec, basedir + 'mea_256.prb')

samuelgarcia:
I am agree with the first implementation but it do not work for several reason just now:
  * the assertion is the wrong way
  * in loadProbeFile ```if isinstance(prop_val, dict) and len(prop_val.keys()) == channels_in_group:``` is 
    not always the case. It is very common that a channel is remove from the 'channels' but not from 
   'geometry'. Sor ```len(prop_val.keys()) == channels_in_group``` is False and we lose the location.

samuelgarcia:
Could you get some file on zenodoo 
https://zenodo.org/record/1205233

https://zenodo.org/record/1205233
```python
basedir = '/media/samuel/SamCNRS/DataSpikeSorting/pierre/zenodo/'


working_folder = basedir + 'run_comparison'

names = ['20160426_patch3']
sorter_list = ['tridesclous', 'spykingcircus']

if os.path.exists(working_folder):
    shutil.rmtree(working_folder)
    
recordings = {}
for name in names:
    dirname = basedir + name + '/'
        
    for f in os.listdir(dirname):
        if f.endswith('.raw') and not f.endswith('juxta.raw'):
            raw_filename = dirname + f
    print(raw_filename)

    with open(raw_filename.replace('.raw', '.txt'), mode='r') as f:
        offset = int(re.findall('padding = (\d+)', f.read())[0])
    print(offset)
       
    rec = se.BinDatRecordingExtractor(raw_filename, 20000, 256, 'uint16', offset=offset, frames_first=True)

        # FOR DEBUG
        rec = se.SubRecordingExtractor(rec, start_frame=0, end_frame=196608)
        
    # HERE THE BUG
        rec = se.loadProbeFile(rec, basedir + 'mea_256.prb')
        
        recordings[name] = rec
    
 run_sorters(sorter_list, recordings,  working_folder, grouping_property='group', debug=True)

```

alejoe91:
I see. I suggest keeping the subrecording behavior and and remove the assertion.
Instead of the assertion we can have a print message: "probe channels are less than recording channels" and "recording channels are less than probe channels". Something like that.

Accordingly we can soften the checks :)

Do you think you can do it?

samuelgarcia:
OK for removing the assertion.

Can I also remove the ```len(prop_val.keys()) == channels_in_group```.
This make the location not written in the new sub PRB file.

alejoe91:
Sure. We should keep it flexible. Maybe also here adding a print message if the location length are not the same

samuelgarcia:
When I read the error of the assertion, I still think that it was in the wrong direction.
So this new proposal is OK with the message.

samuelgarcia:
There is another problem with the flat design of theses properties by channel:
If you want channel to belong to two group with distinct geometry inside each group.
The PRB file handle it correctly, but with the flat design (one channel one group one geometry) you can't.

It is not a usual case but it could be if you want overlapping sorting along probe geometry.

alejoe91:
I think that having the same channel belong to multiple groups might be implemented. But for now let's open an issue about it as it would require to change both the loadProbeFile, saveProbeFile and extractSubExtractorByProperty..

Ok?

alejoe91:
@samuelgarcia is this ready to merge?

samuelgarcia:
Yep.
Ready to merge.

PR #96, opened 2019-03-01
Fixes in MEArec and Biocam recording extractors

mhhennig:
- MEArec now uses memmappped raw data access (via h5py)
- minor Biocam fixes


mhhennig:
Comments on travis build failures:
- The MEArec test creates a directory structure, a feature I removed. Test needs to be adapted to use a hdf5 file instead.
- Biocam files are not correctly created, and we may drop support for writing these files anyway. Should this test be removed?

mhhennig:
Ok, the biocam tests now pass.
The remaining exception comes from the MEArec test. Can you fix it Alessio?

alejoe91:
@mhhennig Ok I'll merge and fix the MEArec load/save functions.

PR #98, opened 2019-03-07
Locations and groups added to extractors

colehurwitz31:


colehurwitz31:
Added locations and groups to the extractors. Please let me know if this is what we had in mind and then I can merge. I tested it and it works from what I can tell. @alejoe91

Issue #104, opened 2019-03-18
writeBinaryDatFormat should not have by default float32


samuelgarcia:
writeBinaryDatFormat force by default float32.
Many big files are int16.
I think we should have by default ```dtype=None``` which would means keep the original dtype.

Issue #103, opened 2019-03-18
**writeBinaryDatFormat** should have transpose=True by default


samuelgarcia:
recording.getTraces().shape give (nb_channel, nb_sample)
but across many soft the raw binary file aka (.dat, .raw, .bin, ...) is internally (nb_sample, nb_channel) so by default writeBinary should be the standard way and so transpose=True.

My guess is that there is a bug in klusta and kilosort around that.

The very best would to NOT have default parameters for this function and force each piece of code to explicite transpose or not.

And the very very best would be to have a more explicit kargs like time_axis=0 or time_axis=1 so the transpose would be infered from that avoiding mistake.

PR #105, opened 2019-03-18
Improve writeBinaryDatFormat:

samuelgarcia:
* dtype=None by default
  * chunksize copy when chunksize is not None

This is about theses problems:
#103 #104 
https://github.com/SpikeInterface/spiketoolkit/issues/72

### DIFF
--- a/spikeextractors/tools.py
+++ b/spikeextractors/tools.py
@@ -180,7 +180,7 @@ def saveProbeFile(recording, probe_file, format=None, radius=100, dimensions=Non
         raise NotImplementedError("Only .csv and .prb probe files can be saved.")
 
 
-def writeBinaryDatFormat(recording, save_path, transpose=False, dtype=None, chunksize=None):
+def writeBinaryDatFormat(recording, save_path, time_axis=0, dtype=None, chunksize=None):
     '''Saves the traces of a recording extractor in binary .dat format.
 
     Parameters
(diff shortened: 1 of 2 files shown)


colehurwitz31:
Yes, these changes are good. We can merge when it is ready.

colehurwitz31:
I also think that explicit kargs like time_axis=0 or time_axis=1 would be good for this and would make it more intuitive.

samuelgarcia:
OK, so I will do the change time_axis instead of transpose.
Wait before the merge.

@alejoe91 : about the possible bug in klusta and kilosort ?

samuelgarcia:
Oups it appear that I was doing a transpose when tranpose=True but in fact this is the reverse.

alejoe91:
maybe `transpose=True` by default?

alejoe91:
or time_axis=1

samuelgarcia:
The correct way for theses files was transpose=False and so time_axis=0.

I really would vote for the default behavior to write the file in the correct order.

alejoe91:
when transpose was False the files were actually transpose though

samuelgarcia:
I don't get it.

What I anderstand is transpose=False the getTrace() are transposed so files are in the correct way (sample, channels).

alejoe91:
Correct :) but now the data are transposed only if transpose=True. So I suggest using transpose=True by default, as in klusta and kilosort the transpose arg is not used

alejoe91:
Ooook! My bad. time_axis=0 is the correct (default) behavior. Sorry for the confusion ;)

alejoe91:
Ooook! My bad. time_axis=0 is the correct (default) behavior. Sorry for the confusion ;)

samuelgarcia:
I change transpose to time_axis to make it less ambiguous for all of us.

If you are happy with this you can merge.

I need this for a refactoring in spiketoolkit.

alejoe91:
good for me :)

Merging

samuelgarcia:
Thanks.

PR #115, opened 2019-03-20
jump to next channel group if no clusters in channel group

lepmik:


colehurwitz31:
Unless I am missing something, won't this have the same behavior as before?

The try catch block just continues if there are no clusters and then it still tries to do the for loop

lepmik:
Is it not continuing to next iteration in the current loop - skipping whatevers beneath?

alejoe91:
@colehurwitz31 @lepmik Yes the continue stops the current iteration and continues to the next. This should fix the problem :)

colehurwitz31:
Ah yes, my bad, just tested it

colehurwitz31:
I will merge

PR #119, opened 2019-03-29
Added extractor list, installation information, gui_params, naming, and installation messages.

colehurwitz31:
Planning to do the same for preprocessing and postprocessing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Build Status](https://travis-ci.org/SpikeInterface/spikeextractors.svg?branch=master)](https://travis-ci.org/colehurwitz31/spikeextractors)
 
 Alpha Development
-Version 0.3.1
+Version 0.4.0
 
 
 # spikeextractors
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ pkg_name="spikeextractors"
 
 setuptools.setup(
     name=pkg_name,
-    version="0.3.1",
+    version="0.4.0",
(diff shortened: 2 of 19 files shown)


colehurwitz31:
@samuelgarcia I tried to copy your sorterlist format, but my extractorlist is not being recognized as a module. Could you possibly figure out what is going on?

alejoe91:
@colehurwitz31 where is the extractorlist file?

colehurwitz31:
Omg, I don't think I uploaded it... that explains it

colehurwitz31:
Feel free to test out the installation and gui_params features to see if they give you the expected results. I will do the same for preprocessing and postprocessing if you guys think this is the correct way to proceed.

colehurwitz31:
Something tricky is how we handle non-simple types like lists or arrays. I think we have to pass them in as a string in the GUI and then parse them into the more complex types.

colehurwitz31:
I think I will merge soon. @samuelgarcia Do you think that this method of installation and gui_params is the right way forward for now? If so, I can implement this in preprocessing/postprocessing in spiketoolkit.

samuelgarcia:
Very few times this week.
Sorry.
This looks good.
have you started UI with pyqtgraph parameters ?
have you tried ephyviewer ?

colehurwitz31:
I looked at ephyviewer and left a comment about what I was confused about. I think I would need you to walk me through the purpose of the different viewers and what kinda feedback you are looking for.

PR #121, opened 2019-04-08
Changed the type information to be more informative. 

colehurwitz31:
@samuelgarcia I am changing the type information in the params and moving away from PyQtGraph types which can only be 'str', 'bool', etc. 

The reasons are as follows:

1. There is an annoying bug for PyQtGraph default values (hasDefault always returns True)

2. I want to include more interesting types like 'list' and 'np.dtype' and 'RecordingExtractor' so I think it is best if we have our own parameter reader. It doesn't need to be as complex as PyQtGraph, but it should handle lists and other types more seamlessly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Build Status](https://travis-ci.org/SpikeInterface/spikeextractors.svg?branch=master)](https://travis-ci.org/colehurwitz31/spikeextractors)
 
 Alpha Development
-Version 0.4.0
+Version 0.4.1
 
 
 # spikeextractors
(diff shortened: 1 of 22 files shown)


colehurwitz31:
@samuelgarcia  There is an annoying bug in PyQtGraph about default values and also I want to include more interesting types like 'list' and 'np.dtype' so I think it is best if we have our own parameter reader. It doesn't need to be as complex as PyQtGraph, but it should handle lists and other types more seamlessly.

samuelgarcia:
In my experience, making our own parameters system is lot of work.
The parametes system of pyqtgraph si rather simple and limited : it is why I like it.

Feel free to do what you want and need about this gui_params.

I would advice to not make too complicated UI. I mean not complicated for the user and not complicated for the code.
Try simple UI and interact with potential user to get feedback before making thing too complex.

I made a lot of UI in the past for ephy experimentalist. Few of then are really usefull :(
So now I try to spend a less as possible my time Qt5...

colehurwitz31:
I agree that making our own parameter system is a lot of work, but all we need these parameter types for is the GUI which is making sure the user enters in the right thing. We don't need any more complexity than that.

PR #123, opened 2019-04-19
Added an option to invert the signal

mhhennig:
Useful if spikes are positive going, and the spike sorter expects negative events.

Question: should every recording extractor do this?


colehurwitz31:
I would call it "invert_signal" for extra clarity, but it seems to make sense to me. I am not sure other extractors will need it because most are format specific.

colehurwitz31:
Although if every extractor returns positive events, then all the spike sorter has to do is invert any extractor. So I am not sure we need the inversion.

alejoe91:
What about adding it directly to the RecordingExtractor? it could be useful in general I guess.

colehurwitz31:
I would be ok with adding that to the base RecordingExtractor with default value False. We would have to implement it separately in each extractor, but that is not a big deal.

magland:
There may be a problem with having a default implementation that default to
False. It is that you can no longer distinguish between the case where it
is just not implemented in the extractor, or whether the value is False. I
think perhaps we should just have properties

mhhennig:
Alternatively we could create a simple function in spiketoolkit.preprocessing? In fact I need a function to invert and scale the data to be able to read and use the kilosort processed data.

Which solution is cleaner? Offer it as part of the extractors, or as part of the pipeline?

colehurwitz31:
I added the HS2 caching that you implemented so I will close this pull request

PR #129, opened 2019-05-01
Implemented SpikeGLX extractor for neuropixels

alejoe91:
- read number of channels
- loads channel locations (assuming standard configuration)

### DIFF
--- a/spikeextractors/extractors/spikeglxextractor/spikeglxrecordingextractor.py
+++ b/spikeextractors/extractors/spikeglxextractor/spikeglxrecordingextractor.py
@@ -1,4 +1,5 @@
 from spikeextractors import RecordingExtractor
+from spikeextractors.tools import read_binary
 import os
 import numpy as np
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


colehurwitz31:
Looks great! Is it possible to memory map the time series? I was wondering if we should do that for the BinDatExtractor also... if it is possible.

alejoe91:
I think it is memmaped. Check out the _read_binary function

colehurwitz31:
I see! That is great. However, we now have to read_binary functions (in SpikeGLX and BinDat). Is that repeated code that we can move to tools.py?

alejoe91:
Yes that's exactly the same, will do it

alejoe91:
done :)

spikeextractors/extractors/spikeglxextractor/spikeglxrecordingextractor.py
```diff
@@ -97,6 +96,9 @@ def _read_binary(file, numchan, dtype, frames_first, offset):
 def _parse_spikeglx_metafile(metafile):
     tot_channels = None
     ap_channels = None
+    x_pitch = 21
+    y_pitch = 20
```
colehurwitz31: Will pitch always be the same here?

colehurwitz31: Is this only for Neuropixels?

alejoe91:
yes for the neuropixels probe, but I could add optional args

colehurwitz31:
If it is possible, we want the SpikeGLX extractor for any probe geometry, right? I may be misunderstanding what SpikeGLX encapsulates.

alejoe91:
SpikeGLX only works with Neuropixels

colehurwitz31:
Got it. So the num channels will always be 385? Or could they use a portion of the channels?

alejoe91:
no it depends on the sync channels. It gets it from the metadata

colehurwitz31:
Ok, so I guess the only change I would recommend is fixing the title of y_pitch in the gui params. Then I will merge. Awesome work Alessio!

PR #138, opened 2019-05-13
Extractor fixes for biocam and hs2

mhhennig:
* better reading performance for Biocam files
* should now correctly detect different brw formats


alejoe91:
is this ready to be merged?

colehurwitz31:
Is the invert data removed from the bindat extractor

mhhennig:
Yes, the invert is removes. For consistency I have decided we should use the preprocessing modules for this.Hee's a request, can we have, instead of the one that scales the data, a 'transformer', that does a*x+b, where a and b are arguments. This will be useful in quite a few cases where the data scale is off for the integer range HS2 expects.

mhhennig:
Looks ready to me...

PR #144, opened 2019-05-29
Proposal for sample rate in sorting extractor

samuelgarcia:


samuelgarcia:
This is just abase for discussion to integrate sample_frequency optionally in sortingextractors.

I made proof of concept in tridesclous.
this idea is totally simple each extractor can ini**\_\_init()\_\_** ceeate the attribute self._sampling_frequency.

What do you think ?

colehurwitz:
I think that can work. Is sampling frequency the only thing we want to add that is similar to this or will we want to add other properties in the future? This is not easily extendable to new properties.

samuelgarcia:
I don't known.
If we adopt this, wa can change SortingComparison fro sample to time soon.
Lets wait @alejoe91 coments.

samuelgarcia:
@alejoe91 @colehurwitz : if you would validate this PR, we could quickly fix this https://github.com/SpikeInterface/spiketoolkit/issues/128

No ?

colehurwitz:
I think the PR looks good. Once the checks pass, we can merge. I don't have a better idea of how to handle sampling frequency.

The only other option I can think of is to have a function like this:

```
def get_sampling_frequency(self):
      pass
```

That can be overwritten by the child classes.. Not sure if that is better at all though.

alejoe91:
probably we could have in the SortingExtractor init:

`self._sampling_frequency = None`

and

```
def get_sampling_frequency(self):
      self._sampling_frequency
```

So everything we need to do is initialize `self._sampling_frequency` in each extractor.

Good idea? :)

samuelgarcia:
@alejoe91 : have a look at files. What you say is exactly the proposal.

@colehurwitz : I think I prefer the solution I proposed because it will be more or less one line to add for each extrator.

alejoe91:
@samuelgarcia exactly :) 

Let's keep this open and add the same for the other extractors. I will have some time on Tuesday to do it

samuelgarcia:
What about merging this and make other extractor later on.

alejoe91:
Why is the test failing? I'll try to add some today. Also we should have recording extractors for all sorting extractors, as the information is usually there

alejoe91:
ok I'll merge this and then work on the other extractors on a separate PR.

@samuelgarcia is there a specific reason why you PR from you fork rather than a branch? It would be easier to work together on the same PR is done from the branch in my opinion

samuelgarcia:
@alejoe91 : it was from a branch : new_features. I think it is better if anyone works in its own forks for PR. No ?

alejoe91:
I disagree actually because if I want to make some changes I have to fork a your fork and make a PR to your branch..do you agree?

samuelgarcia:
There is a scenario for that : 
   * fork from the main repo in alesio
   * git clone from alesio repo
  * add remote upstream from the main repo (spikeinterface)
  * add remote from samuelgarcia
  * checkout my branch from my repo
  * make a (sub) PR on my repo
  * I merge your PR in my repo
  * and then merge the full PR on the main repo

This sounds crazy but this is a better way I think...

What you did is fork from my repo. I think it is a mistake.

alejoe91:
I see. But what is the benefit of this instead of using a branch? I don't know!

samuelgarcia:
I think working directly on a main team repo is a bad pratice even in branch.
I don't known excatly why, I read it long time ago and it is now very intuitive to me to not do it.

PR #146, opened 2019-05-31
syntax for import spikeextractors typo correction

JoseGuzman:
in 'from spikeextractors import se', must be 'import spikeextractors as se'


alejoe91:
Thanks @JoseGuzman 

We are improving the documentation at the moment, including read the docs for all repos.
We should have something more complete in 2-3 weeks :)

Cheers
Alessio

PR #162, opened 2019-06-11
Check if 'plane' in recgen.info['electrode'] dict

alejoe91:


colehurwitz:
I think the issue is that if the user sets "locs_2d=True", they shouldn't end up with 3D locations (if they don't have the 'plane' info in their file).

Maybe, by default, it should pick [y,z] if they don't have that info.

alejoe91:
if the plane is not there then you can't really know which are the default locations, ad the user should add it manually. Which probe is the problem for?

colehurwitz:
This was an issue for a recent recording generated for 'SqMEA-10-15um'. It did not have the 'plane' information. 

However, setting "locs_2d=True", and then getting 3D locations (because there is no 'plane'), doesn't make sense. It should just fail then I think and say that there isn't enough info to get 2D locations. It shouldn't silently ignore your request for 2D locations.

alejoe91:
@colehurwitz now if plane is not found it uses yz and prints a message

colehurwitz:
Awesome. That is even better than what I had in mind.

PR #163, opened 2019-06-11
Made get unit ids return a numpy array in CurationSortingExtractor

colehurwitz:


colehurwitz:
Any issues with this? I will merge this since i don't think it should be a problem

samuelgarcia:
I thougth the API was a list.
Some extractor return a list.
It should be always teh same and tested.

colehurwitz:
Sorry for merging this!!

> I thougth the API was a list.

The API says that it can be either a list or a 1D numpy array. I would argue that either works, but we should choose one.

colehurwitz:
```
@abstractmethod
    def get_unit_ids(self):
        '''This function returns a list of ids (ints) for each unit in the recording.
        Returns
        ----------
        unit_ids: array_like
            A list or 1D array of the unit ids in recording (ints).
        '''
        pass
```

colehurwitz:
I am making an issue to discuss. Sorry for merging, I didn't think anyone was gonna read this small change, but I was wrong.

PR #167, opened 2019-06-11
Fix NpzSortingExtractor.write_sorting when empty.

samuelgarcia:


samuelgarcia:
spiketoolkit bug because of this.

PR #173, opened 2019-06-12
Added sampling frequency and RecordingExtractors for all sorters

alejoe91:
+ some code cleaning


colehurwitz:
This covers all the sorting extractors that have access to this info? Which formats don't have access?

spikeextractors/extraction_tools.py
```diff
@@ -33,6 +33,27 @@ def read_python(path):
...
+        for k, v in dict.items():
+            if isinstance(v ,str) and not v.startswith("'"):
+                if 'path' in k and 'win' in sys.platform:
+                    f.write(str(k) + " = r'" + str(v) + "'\n")
+                else:
+                    f.write(str(k) + " = '" + str(v) + "'\n")
+            else:
+                f.write(str(k) + " = " + str(v) + "\n")
+
+
```
samuelgarcia: Why do we need a home made seriliazer ?

spikeextractors/extractors/phyextractors/phyextractors.py
```diff
@@ -232,10 +242,13 @@ def write_sorting(sorting, save_path):
         spike_times = spike_times[sorting_idxs]
         spike_clusters = spike_clusters[sorting_idxs]
 
+        params = {'sample_rate': sorting.get_sampling_frequency()}
         if not save_path.is_dir():
-            save_path.mkdirs()
-        np.save(save_path /'spike_times.npy', spike_times[:, np.newaxis].astype(int))
+            save_path.mkdir(parents=True)
+        write_python(save_path / 'params.py', params)
+        np.save(save_path / 'spike_times.npy', spike_times[:, np.newaxis].astype(int))
```
samuelgarcia: Personnaly I prefer to be totally explicit on dtpe and put 'int64'

alejoe91:
@colehurwitz Trideclous is missing.

@samuelgarcia can you add the TridesclousRecoridngExtractor?

spikeextractors/extraction_tools.py
alejoe91: this takes care of windows paths. Feel free to change it! but we should test on windows as well

spikeextractors/extractors/phyextractors/phyextractors.py
alejoe91: I can make it int64

spikeextractors/extractors/spykingcircusextractors/spykingcircusextractors.py
```diff
@@ -0,0 +1,154 @@
...
+                        result_folder = spykingcircus_folder
+
+        assert isinstance(parent_folder, Path) and isinstance(result_folder, Path), "Not a valid spyking circus folder"
+
+        for f in parent_folder.iterdir():
+            if f.suffix == '.params':
+                sample_rate = _load_sample_rate(f)
+            if f.suffix == '.npy':
+                recording_file = str(f)
+        NumpyRecordingExtractor.__init__(self, recording_file, sample_rate)
```
samuelgarcia: Are you sure that all .npy in SC are traces always ?

samuelgarcia:
@alejoe91 : I can do a tridesclous recording extractor but not before the meeting.
I run out of times.

spikeextractors/extractors/spykingcircusextractors/spykingcircusextractors.py
alejoe91: No, but the our wrapper uses .npy. I'll leave as it is for now :)

spikeextractors/extraction_tools.py
alejoe91: this takes care of windows path. Feel free to modify it if you have a better solution

colehurwitz:
I am a little confused. What is the purpose of adding recording/sorting extractors that just call other recording/sorting extractors. eg. Kilosort calling Phy extractor.

alejoe91:
Kilosort saves its output in phy format

colehurwitz:
Yes, by why having both extractors? Couldn't we just have the Phy extractors?

alejoe91:
then we have to change the sorters in spiketoolkit

colehurwitz:
I mean, I don't mind having both extractors, just not sure the purpose.

alejoe91:
I think that if someone runs kilosort then it's more straightforward to have a Kilosort extractor.
The user might not know that it's the same as the phy format IMO

colehurwitz:
Ok, I am fine with merging

PR #176, opened 2019-06-13
Bug in npz when no sampling_rate

samuelgarcia:
This bug make the comparison study broken if sampling frequency is unkonwn.


colehurwitz:
Good catch!

samuelgarcia:
Ok I merge.

colehurwitz:
I think I have to go do this for HS2. Not sure about the other formats

PR #177, opened 2019-06-13
Set to None if sampling not in folder

colehurwitz:
@alejoe91 Do any of the formats you added have this proble? Samuel made me realize that since write_sorting doesn't always write out the sampling frequency, we need to check to see if it is in the file.


alejoe91:
The self._sampling_frequency is initialized to none in the base SortingExtractor. So you don't really need the else. I think all the sorters that I did check for that

samuelgarcia:
@alejoe91 : agree but it is better to be explicit, this make the code readable easily.

PR #178, opened 2019-06-14
Fix to read prb files that contain python 2 usage of range()

mhhennig:


colehurwitz:
I think it looks good, but @alejoe91 understands probe logic better than I do

alejoe91:
yeah this is only the read_python function :) merging!

PR #183, opened 2019-06-18
parse dtype as str (for numpy dtypes)

alejoe91:


samuelgarcia:
I am not sur at all this is a good idea.
Why are you doing this ?

alejoe91:
because it fails if the type is a numpy.dtype. The startswith doesn't work

alejoe91:
and it was breaking our entire pipeline :P

samuelgarcia:
And now it is breaking mine ;)

alejoe91:
Oh no!! ok sorry about that :) What can we do then? maybe check uint only if not isinstance(self._dtype, numpy.dtype)? 

How is it breaking your code?

colehurwitz:
So the dtype is now a string? Like 'int' or something? What are the options that can be passed in for np.dtypes. Having it as a string will be useful for the GUI since strings are easiest

samuelgarcia:
@alejoe91 it was a joke.

samuelgarcia:
I think the good pratice is to do:
```python
def my_fun(self, dtype):
    self._dtype= np.dtype(dtype) # so  always a dtype

# and make test like this

if self._dtype.str.startswith(...)
```

samuelgarcia:
You also have dtype.kind.

real dtype have convinient methods.

samuelgarcia:
But no worry I don't care at the end.

alejoe91:
ahahahah I took it seriously. `self._dtype.str.startswith(...)` this looks good!

PR #187, opened 2019-06-20
Fixes to h5 order in MEArec and KlustaExtractor

alejoe91:
- if channel_ids is not ordered in MEArecRecordingExtractor it throws an Error. Patched it
- if running klusta without grouping property but the group information is available, klusta still runs split by group and the unit_ids must be reset to have unique ids


colehurwitz:
Why does this throw an error? A numpy array can have an unordered channel_ids passed in so I don't see why this would fail..

alejoe91:
It's h5 as the PR says :)

alejoe91:
can I merge this?

samuelgarcia:
yep

PR #196, opened 2019-06-30
Updated MdaSortingExtractor to deal with sampling freq

alejoe91:
+ fixes to naming and use pathlib


magland:
I see what you did there, but I don't really like it. If you point to a firings.mda, it should not get info from params.json that is next to the firings.mda. Instead, let's pass samplerate=30000 into the constructor of MdaSortingExtractor. It will need to be passed explicitly (and it can be obtained from the accompanying recording extractor)

PR #205, opened 2019-07-04
option to not check suffix for mearec recordings - WIP

magland:
I'd like to be able to tell the mearec extractors NOT to worry about the file name suffix for mearec recordings.


magland:
But we need to wait until mearec also has the check_suffix flag in load_recordings()
@alejoe91

colehurwitz:
@alejoe91 Should I wait for this Alessio?

alejoe91:
https://github.com/alejoe91/MEArec/blob/master/MEArec/tools.py#L254

PR #206, opened 2019-07-04
 All multi extractors are working and can be called with functions.

colehurwitz:
I created two new notebooks to show how to call and use these multiextractors and added some unit tests for them.

se.concatenate_recordings_by_time(recordings, epochs) --> MultiRecordingTimeExtractor
se.concatenate_recordings_by_channel(recordings, groups) --> MultiRecordingChannelExtractor
se.concatenate_sortings(sortings) --> MultiSortingExtractor

One thing I noticed, both the MultiSorting and MultiRecordingChannel extractor change the underlying channel properties of the recordings that they concatenated... that is fine, but when you pass in duplicate recordings or sortings, the set_property calls will overwrite each other... I don't think this is a problem though because we should be expecting different extractors to be passed in.

Let me know if this pull request is ready please! @samuelgarcia @alejoe91 @magland

### DIFF
--- a/spikeextractors/MultiRecordingChannelExtractor.py
+++ b/spikeextractors/MultiRecordingChannelExtractor.py
@@ -1,0 +1,102 @@
+from .RecordingExtractor import RecordingExtractor
+import numpy as np
+
+# Concatenates the given recording channels in time
+
+class MultiRecordingTimeExtractor(RecordingExtractor):
+    def __init__(self, recordings, epoch_names=None):
+        RecordingExtractor.__init__(self)
+        if epoch_names is None:
+            epoch_names = [str(i) for i in range(len(recordings))]
+
+        # Add all epochs to the epochs data structure
+        start_frames = 0
+        for i, epoch_name in enumerate(epoch_names):
+            num_frames = recordings[i].get_num_frames()
+            self.add_epoch(epoch_name, start_frames, start_frames + num_frames)
+            start_frames += num_frames
+
+        self._RXs = recordings
+
+        # Num channels and sampling frequency based off the initial extractor
+        self._first_recording = recordings[0]
+        self._num_channels = self._first_recording.get_num_channels()
+        self._channel_ids = self._first_recording.get_channel_ids()
+        self._sampling_frequency = self._first_recording.get_sampling_frequency()
+
+        # Test if all recording extractors have same num channels and sampling frequency
+        for i, recording in enumerate(recordings[1:]):
+            channel_ids = recording.get_channel_ids()
+            sampling_frequency = recording.get_sampling_frequency()
+
+            if (self._channel_ids != channel_ids):
+                raise ValueError("Inconsistent channel ids between extractor 0 and extractor " + str(i + 1))
+            if (self._sampling_frequency != sampling_frequency):
+                raise ValueError("Inconsistent sampling frequency between extractor 0 and extractor " + str(i + 1))
+
+        self._start_frames = []
+        self._start_times = []
+        ff = 0
+        tt = 0
+        for RX in self._RXs:
+            self._start_frames.append(ff)
+            ff = ff + RX.get_num_frames()
+            tt = tt + RX.frame_to_time(0)
+            self._start_times.append(tt)
+            tt = tt + RX.frame_to_time(RX.get_num_frames()) - RX.frame_to_time(0)
+        self._start_frames.append(ff)
+        self._start_times.append(tt)
+        self._num_frames = ff
+
+        # Set the channel properties based on the first recording extractor
+        self.copy_channel_properties(self._first_recording)
+
+    def _find_section_for_frame(self, frame):
+        inds = np.where(np.array(self._start_frames[:-1]) <= frame)[0]
+        ind = inds[-1]
+        return self._RXs[ind], ind, frame - self._start_frames[ind]
+
+    def _find_section_for_time(self, time):
+        inds = np.where(np.array(self._start_times[:-1]) <= time)[0]
+        ind = inds[-1]
+        return self._RXs[ind], ind, time - self._start_times[ind]
+
+    def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_frames()
+        RX1, i_sec1, i_start_frame = self._find_section_for_frame(start_frame)
+        RX2, i_sec2, i_end_frame = self._find_section_for_frame(end_frame)
+        if i_sec1 == i_sec2:
+            return RX1.get_traces(channel_ids=channel_ids, start_frame=i_start_frame, end_frame=i_end_frame)
+        traces = []
+        traces.append(
+            self._RXs[i_sec1].get_traces(channel_ids=channel_ids, start_frame=i_start_frame,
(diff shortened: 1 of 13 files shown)


spikeextractors/MultiRecordingChannelExtractor.py
```diff
@@ -0,0 +1,115 @@
...
+# Concatenates the given recordings by channel
+
+class MultiRecordingChannelExtractor(RecordingExtractor):
+    def __init__(self, recordings, groups=None):
+        RecordingExtractor.__init__(self)
+        self._RXs = recordings
+        self._all_channel_ids = []
+        self._channel_map = {}
+
+        # Sampling frequency based off the initial extractor
```
samuelgarcia: make we could check that sampling rate are all the same here ?

spikeextractors/MultiRecordingChannelExtractor.py
```diff
@@ -0,0 +1,115 @@
...
+        self._first_recording = recordings[0]
+        self._sampling_frequency = self._first_recording.get_sampling_frequency()
+        self._num_frames =  self._first_recording.get_num_frames()
+
+        # Test if all recording extractors have same sampling frequency
+        for i, recording in enumerate(recordings[1:]):
+            sampling_frequency = recording.get_sampling_frequency()
+            if (self._sampling_frequency != sampling_frequency):
+                raise ValueError("Inconsistent sampling frequency between extractor 0 and extractor " + str(i + 1))
+
```
samuelgarcia: Oups sorry this is done here.

spikeextractors/MultiRecordingChannelExtractor.py
```diff
@@ -0,0 +1,115 @@
+from .RecordingExtractor import RecordingExtractor
+import numpy as np
+
+# Concatenates the given recordings by channel
+
+class MultiRecordingChannelExtractor(RecordingExtractor):
+    def __init__(self, recordings, groups=None):
+        RecordingExtractor.__init__(self)
+        self._RXs = recordings
```
samuelgarcia: I would pu self._recordings intstead of self._RXs

spikeextractors/MultiRecordingChannelExtractor.py
```diff
@@ -0,0 +1,115 @@
...
+            if (self._sampling_frequency != sampling_frequency):
+                raise ValueError("Inconsistent sampling frequency between extractor 0 and extractor " + str(i + 1))
+
+        #set channel map for new channel ids to old channel ids
+        new_channel_id  = 0
+        for r_i, RX in enumerate(self._RXs):
+            channel_ids = RX.get_channel_ids()
+            for channel_id in channel_ids:
+                self._all_channel_ids.append(new_channel_id)
+                self._channel_map[new_channel_id] = {'rx': r_i, 'channel': channel_id}
```
samuelgarcia: Same here 'rx' > 'recording'
'channel' > 'channel_id'

spikeextractors/MultiRecordingChannelExtractor.py
```diff
@@ -0,0 +1,115 @@
...
+                new_channel_id += 1
```
alejoe91: Maybe here you could check if the channels have duplicates. For instance, if you concatenated 2 recordings with channels: [0,1,2,3], [8,9,10,11] the channel ids should be maintained

spikeextractors/extractors/tridescloussortingextractor/tridescloussortingextractor.py
```diff
@@ -47,4 +47,4 @@ def get_unit_spike_train(self, unit_id, start_frame=None, end_frame=None):
             spike_times = spike_times[spike_times >= start_frame]
         if end_frame is not None:
             spike_times = spike_times[spike_times < end_frame]
-        return spike_times.copy()  #�copy avoid reference to the unerlying memmap
+        return spike_times.copy()  #�copy avoid reference to the unerlying memmap
```
alejoe91: ?

alejoe91:
Looks good @colehurwitz 

Just had some comments :)

samuelgarcia:
Hi @colehurwitz .
I am not evry familiar yet with MultiRecordingExtractor.
Your PR looks good. 
I made few comments.

Globaly, I don't like the general approach of this MultiRecordingExtractor (time and channel).
This comment is outside this PR.
Because it concatenate in memory evrything. This is won't work for big dataset.
If we need to concatenate, 3 recording of 1 hour the among of memory needed to make this is phenomenal for high channel count.
I think I prefer the explicit approach of multi segment neo like where the user need to loop explicitly on recorings (segment)

colehurwitz:
Hi all! Thanks for the suggestions, I will correct a few things about my PR.

Unless I am mistaken @samuelgarcia, we only ever load things into memory when calling get_traces and even then, we would never load the whole recording in. The MulitRecording extractors just coordinate extraction from multiple files.

samuelgarcia:
@colehurwitz : yes it is true.
I don't known how to explain but I prefer explicit.
This magic auto concatenate do not make happy I prefer API where looping over recording sis explicit and not hidden.
phy/klust/sc have the same approach of magic concatenate and I don't like it so much.
End user of spikeextractors are devs not experimentalist, so explicit loop over segment is no so hard.

Nevertheless, I already pronounce myself on this and this is not related to this PR.
So no worry.

alejoe91:
I agree with Cole. Only when get_traces is called (for the required start_frame and end_frame) the data is actually loaded in memory.

If you want to concatenate to spikesort, you have to dump all the files in the same output file, at least for most sorters

samuelgarcia:
> If you want to concatenate to spikesort, you have to dump all the files in the same output file, at least for most sorters
Because the only sorter that deal with multi segment is tdc :)
All others play with a one segment only.

We could push a layer that would help devs to loop explicitly over segments.
here the magic concatenate is a trap because  at the end of the pipe spiek index are absolut and we need a trick to assign them back to the related file and segment which is for the end and (experimentalist) a pain.

As a end user, to retrieve spike I would prefer a list of sorting that explicitly is the same size of the recording list and not spike indexe that relate to the concatenated recordings.

colehurwitz:
I see your point @samuelgarcia. I think why we liked the auto-magic approach is that spike sorters and tools can just treat the Multi-extractors the same way as normal extractors which makes development a little easier in my opinion :) 

However, the explicit approach might help us avoid issues in the future that I am not yet aware of. It is something to keep in mind for sure!

samuelgarcia:
>  to keep in mind for sure!

Trust i have this in mind for the futur :)

colehurwitz:
We could maybe add a function to the Multi extractor like: ‘get_traces_explicit’ which makes the call an explicit one or something! Something I will think about as well

alejoe91:
I would say let's merge for now and we can implement the `segment` approach later :)

samuelgarcia:
Of course.
Maybe next year. We more urgent thing to do now (5 papers!!)

spikeextractors/MultiRecordingChannelExtractor.py
colehurwitz: I would rather remap I think just to keep the behaviour consistent

spikeextractors/extractors/tridescloussortingextractor/tridescloussortingextractor.py
colehurwitz: I'll fix this in the next pull request I make. Weird!

PR #218, opened 2019-07-31
rmv source from NWB interfaces

bendichter:
'source' args were removed from NWB just prior to the 2.0 release


alejoe91:
@bendichter thanks! ready to merge?

bendichter:
@alejoe91 yes. The NWB extractor might require one or two more adjustments but this will be a step in the right direction.

alejoe91:
Ok merging! Thanks again

PR #219, opened 2019-07-31
make sure samplerate is returned in kHz

bendichter:


alejoe91:
@bendichter we use sampling frequency in Hz in all other extractors. Are the timestamps in s or ms?
In case they are in ms then you change return it in Hz, right?

PR #221, opened 2019-08-04
Fix bug in get snippets

colehurwitz:
get snippets wasn't correct shifted for the start frame


colehurwitz:
In subrecordingextractor

Issue #227, opened 2019-08-09
units of measurement / magnitude


bendichter:
It is common for data acquisition formats to store voltage data as `int16` values and have a scaling factor to convert these values to volts (e.g. SpikeGLX). It looks like spikeextractor ignores these scaling factor values and returns data in an arbitrary magnitude. As a result, the units of measurement are lost during conversion from one data format to another. Would it be possible to include `get_scaling_factor` or something so that we can resolve data in volts?

alejoe91:
@bendichter I think this should be handled by each extractor. I implemented it in the `OpenEphysRecordingExtractor` and you can get traces eithe in int16 or in float (uV) (https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/openephysextractors/openephysextractors.py#L42-L53). I agree we should add something like that in the SpikeGLXExtractor.

Also, the `get_traces` function should allow to return traces both in `int16` or `uV`, as many sorters use `int16` directly.

We could have both a `get_scaling_factor`  and  a `get_dtype`. In case the traces are in uV, then the scaling factor could be one. 

@colehurwitz @bendichter what do you guys think?

bendichter:
@alejoe91 Thanks for pointing out that feature. NWB allows you to have it both ways: store data in `int16` which saves memory and also a conversion factor which allows you to convert to volts (much like the `gain` value of OpenEphys). That's why I thought a `get_scaling_factor` would be particularly useful for us. Is it specified somewhere that the data is converted to uV as opposed to mV or V? If not, can we do that?

@luiztauffer found a SpikeGLX library that converts to volts (https://billkarsh.github.io/SpikeGLX/Support/SpikeGLX_Datafile_Tools.zip), so we can base it off of that.

alejoe91:
@bendichter I'd prefer using uV as spike amplitudes are not e intuitive imo. 
We can add a `return_unit` in the `get_traces` from the base class and we can convert to uV, mV, or V in the get_traces on the fly. How does it sound? If return_unit is None and the file is binary then int16 is returned. We can definitely port some code from the SpikeGLX library. 
I'm a bit busy next week, would you have time to give it a try?

bendichter:
[...]

PR #228, opened 2019-08-16
add get_conversion_factor

bendichter:
to RecordingExtractor, NumpyRecordingExtractor, and NwbRecordingExtractor

fix https://github.com/SpikeInterface/spikeextractors/issues/227

### DIFF
--- a/spikeextractors/RecordingExtractor.py
+++ b/spikeextractors/RecordingExtractor.py
@@ -80,6 +80,17 @@ class RecordingExtractor(ABC):
         pass
 
     @abstractmethod
+    def get_conversion_factor(self):
+        '''Returns the number you multiply by to convert trace data into uV
+
+        Returns
+        -------
+        conversion_factor: float
+            number you multiply by to convert trace data into uV
+        '''
+        return
+
+    @abstractmethod
     def get_channel_ids(self):
         '''Returns the list of channel ids. If not specified, the range from 0 to num_channels - 1 is returned.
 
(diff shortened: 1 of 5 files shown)


colehurwitz:
My recommendation would be to make conversion factor=1 by default (in RecordingExtractor) and then have get_conversion_factor return that as normal. 

Then, we can override specific conversion factors in the given extractors as needed, but by default, all the tests will pass.

colehurwitz:
Also, the spikeglx changes look good :)

bendichter:
We have an issue with spikeglx. All of the channels have unique gains.

colehurwitz:
We could have gain be a channel property that is loaded in then during initialisation.

alejoe91:
We could have the conversion factor be either a scalar (if the same for all channels) or an array. In the second case we can load it as a property and use channel-specific gains in the get_traces

colehurwitz:
I would load the scalar as a channel property for each channel as well potentially, just to be consistent. Then we can do the same procedure in get traces regardless of the gain being a scalar or an array.

bendichter:
@colehurwitz @alejoe91 I hadn't thought of using the electrode properties. I think of gain more as a property of the dataset than a property of the channel. That approach wouldn't work for NWB:N because electrodes might be associated with multiple datasets that could potentially have different gains. Making `gain` a property of each electrode could work in your system if you are sure that you'll only have on dataset per channel, and doing it this way in SI and storing it differently in NWB:N would be fine.

colehurwitz:
@bendichter @luiztauffer So I implemented the conversion factor already, but the SpikeGLX changes are still very useful. Could one of you make a pull request with just the spikeglx changes?

colehurwitz:
Just enquiring about the spikeglx question above. Can we still use that and forget about the other part of the pull request?

bendichter:
Yeah that's the plan, we just had some pressing issues that took priority
over the last week. We'll get back to this over the next week

PR #231, opened 2019-08-22
Added gain setting and getting

colehurwitz:
@alejoe91 @bendichter 

Thoughts on setting and getting gain? I just implemented it in the same way we handle locations and groups (as a property that can be added and returned). In the initialization of the NWBRecordingExtractor, you can set the gains using this function. 

If you put a float/int, it will set all electrode gains to the same value
If you put a list/array_like of floats/ints, then it will set each electrode according to the list you gave as long as the lengths are the same.

### DIFF
--- a/spikeextractors/MultiRecordingChannelExtractor.py
+++ b/spikeextractors/MultiRecordingChannelExtractor.py
@@ -79,20 +79,12 @@ class MultiRecordingChannelExtractor(RecordingExtractor):
         channel_id_recording = self._channel_map[channel_id]['channel_id']
         return recording.get_channel_property(channel_id_recording, property_name=property_name)
 
-    def get_channel_property_names(self, channel_id=None):
-        if channel_id is None:
-            property_names = []
-            for recording in self._recordings:
-                property_names_recording = recording.get_channel_property_names()
-                for property_name_recording in property_names_recording:
-                    property_names.append(property_name_recording)
-            property_names = sorted(list(set(property_names)))
-        else:
-            recording = self._recordings[self._channel_map[channel_id]['recording']]
-            channel_id_recording = self._channel_map[channel_id]['channel_id']
-            property_names = recording.get_channel_property_names(channel_id_recording)
+    def get_channel_property_names(self, channel_id):
+        recording = self._recordings[self._channel_map[channel_id]['recording']]
+        channel_id_recording = self._channel_map[channel_id]['channel_id']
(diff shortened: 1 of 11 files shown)


colehurwitz:
If you try to get the gain when it has not been set (it is unknown maybe), then it will throw an error. Is this not the default behaviour you want?

alejoe91:
I think it's ok to set it to None by default. We could have a self._gain_set flag, for example.

If the gain is None, then the traces are returned as they are.

bendichter:
I'd rather have get_gain return None. That will have the same effect but will also allow a writer to test for that case and gracefully handle it

alejoe91:
Sounds good to me. Then in the get_traces we can just check if the `gain` property exist.

colehurwitz:
So I decided to set the gain property by default to None. Subclasses can overwrite this if they want with real gain values.

alejoe91:
hy do we have to add the gain property if it's None? I think it's fair to set it only if the set_gain is called with a meaningful value

colehurwitz:
So here is the issue. If we let `get_gain` return None when there isn't a gain property, then that breaks the system we have in place where properties that don't exist in the extractor have an error thrown when someone queries them. Rather than breaking our current system, we need to adhere to the rules and add it is a property that is None by default.

colehurwitz:
So we either need to change the rules and have all properties return None when they are not set or we have to keep the same rules about error throwing and set some properties to None if we want them to be returned as None by default.

colehurwitz:
@bendichter @alejoe91 I honestly think that get_gains should not return None, but rather you should check to see if the property exists for the unit or not. Returning None is often dangerous as it gives the illusion that the value is meaningful. You can gracefully check to see if the property exist or not before converting.

bendichter:
I can work with that. The main point is that I don't want the API to create the impression that data is properly scaled when it is not. I would worry about making the gain default to 1 because it implies that the data is already scaled to uV which may or may not be the case. We'll have to look at each recording format and make that determination, but before we do each one should be stored in some way that indicates that scaling has not yet been dealt with. If we do that by omitting the gain and checking if it is missing that would be OK with me.

bendichter:
fyi @luiztauffer

colehurwitz:
@bendichter So I added a function to check to see if a subset of channel ids (or all channel ids) share a property. If they do, you can grab them all at once like this:

```
if 'gain' in recording.get_shared_channel_property_names(channel_ids):
     gain = recording.get_channel_gains(channel_ids)
```

PR #236, opened 2019-08-28
Rename without upper case for file to avoid colision with class name.

samuelgarcia:
This is needed for spikeinterface metapackage.


samuelgarcia:
dir(spikeinterface.extractors) and dir(spikeextractors) still have differents length.

This is strange.

alejoe91:
@samuelgarcia you can fix the tests by changing 
`from .SubRecordingExtractor`->`from .subrecordingextractor` in line 576 of recordingextractor.py
`from .SubSortingExtractor`->`from .subsortingextractor` in line 567 of sortingextractor.py

I tried to push from this PR but it creates a new branch...

Does this fix the import error of spikeinterface?

samuelgarcia:
Hi Alessio, I will fix test tomorow.
No it do not fix the spikeinterface flat or not flat not everything is in the metapackage namespace.
I don't anderstand why, I will have a deeper look.
It could be a loop import somewhere.

samuelgarcia:
normally you can push in my branch when it is a PR di you add the samuelgarcia/spikeextractors as remote ?

samuelgarcia:
This pass test.
I think we can merge.
This do not fix the flat import but it is better this way.
no ?

alejoe91:
Yes!

PR #245, opened 2019-09-06
Added shared unit and feature name functions

colehurwitz:
Added new tests as well for the shared name functions as well


colehurwitz:
I will merge soon since everything is passing.

PR #248, opened 2019-09-09
added internal extraction tools

colehurwitz:
Based off the meeting we had, I implemented extraction tools that are _internal_ to the RecordingExtractor and Sorting Extractor. Now you can call load_probe_file, save_to_probe_file, write_to_binary_dat_format, get_sub_extractors_by_property with:

```
recording.load_probe_file
recording.save_to_probe_file
recording.write_to_binary_dat_format
recording.get_sub_extractors_by_property
```
You can also call:

```
sorting.get_sub_extractors_by_property
```
I left the original functions in extraction_tools and the extractors call those original function internally. I added tests for them and they pass.

I changed two names as well:

save_probe_file --> save_to_probe_file
write_binary_dat_format --> write_to_binary_dat_format

### DIFF
--- a/spikeextractors/__init__.py
+++ b/spikeextractors/__init__.py
@@ -9,7 +9,7 @@ from .multisortingextractor import concatenate_sortings, MultiSortingExtractor
 from .extractorlist import *
 
 from . import example_datasets
-from .extraction_tools import load_probe_file, save_probe_file, read_binary, write_binary_dat_format, \
+from .extraction_tools import load_probe_file, save_to_probe_file, read_binary, write_binary_dat_format, \
     get_sub_extractors_by_property
 
 from .version import version as __version__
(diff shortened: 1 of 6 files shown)


colehurwitz:
@alejoe91

colehurwitz:
I think we should use this convention throughout spikeinterface.

alejoe91:
Hi @colehurwitz, for now let's keep both. Also can you make a PR to spikesorters? This will break several of them.

I think that we should also port the `write_to_binary` to the `write_recording` of the `BinDatRecordingExtractor`. The correct implementation is the `write_to_binary`, but I think its place should be there

colehurwitz:
We can keep both for now. I will switch to the object version for a bunch of the code though (and maybe the examples).

I kinda like having a write_to_binary in the RecordingExtractor as it is easier to call that then to import a binary extractor and then write. Maybe we keep both? Not sure

colehurwitz:
I'll merge for now and we can either keep/remove the write_to_binary later

PR #254, opened 2019-09-12
Important changes to file API (mode, is_writable, gui_params)

colehurwitz:
Important changes to file API:

Added:

- 'is_writable'
- 'mode' (file, dir, file_or_dir, custom) --> Custom currently covers one case that is unique (neuroscope). 
-- Arguments (based on mode): file --> file_path, dir --> dir_path, file_or_dir --> file_or_dir_path, custom --> anything
- Fixed gui params

@alejoe91 @samuelgarcia If you could check each extractor to make sure what I put makes sense, that would be very helpful. It was a tedious task and I may have make a mistake.


alejoe91:
Looks good Cole! Couldn't spot any mistakes from the commit

PR #264, opened 2019-09-24
Cache extractor

alejoe91:
Implemented a CacheRecordingExtractor. It takes a recording as input and it inherits from a BinDatRecordingExtractor.

When the object is deleted the temporary file is removed

### DIFF
--- a/spikeextractors/cacherecordingextractor.py
+++ b/spikeextractors/cacherecordingextractor.py
@@ -1,0 +1,18 @@
+from spikeextractors.extractors.bindatrecordingextractor import BinDatRecordingExtractor
+import tempfile
+
+
+class CacheRecordingExtractor(BinDatRecordingExtractor):
+
+    extractor_name = 'CacheRecordingExtractor'
+
+    def __init__(self, recording, dtype=None, chunksize=None):
+        tmp_file = tempfile.NamedTemporaryFile(suffix=".dat").name
+        if dtype is None:
+            dtype = recording.get_traces(start_frame=0, end_frame=2).dtype
+        recording.write_to_binary_dat_format(save_path=tmp_file, dtype=dtype, chunksize=chunksize)
+        BinDatRecordingExtractor.__init__(self, tmp_file, numchan=recording.get_num_channels(),
+                                          recording_channels=recording.get_channel_ids(),
+                                          sampling_frequency=recording.get_sampling_frequency(),
+                                          dtype=dtype)
+        self.copy_channel_properties(recording)
--- a/spikeextractors/cacherecordingextractor.py
+++ b/spikeextractors/cacherecordingextractor.py
@@ -1,5 +1,6 @@
 from spikeextractors.extractors.bindatrecordingextractor import BinDatRecordingExtractor
 import tempfile
+import os
 
 
 class CacheRecordingExtractor(BinDatRecordingExtractor):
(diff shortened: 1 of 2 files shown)


spikeextractors/cacherecordingextractor.py
```diff
@@ -0,0 +1,22 @@
...
+import os
+
+
+class CacheRecordingExtractor(BinDatRecordingExtractor):
+
+    extractor_name = 'CacheRecordingExtractor'
+
+    def __init__(self, recording, dtype=None, chunksize=None):
+        self._tmp_file = tempfile.NamedTemporaryFile(suffix=".dat").name
+        if dtype is None:
```
samuelgarcia: I think we should not have the possibility to change this dtype. It should be always the same as the input.

samuelgarcia:
Hi Alessio, this is cool.
Not happy that dtype can be changed, it is dangerous.

By the way. I always wanted that dtype would be part of the extarctor API like recording.get_dtype()
Could we put the trick at class level to get the dtype ?

spikeextractors/cacherecordingextractor.py
```diff
@@ -0,0 +1,22 @@
...
+            dtype = recording.get_traces(start_frame=0, end_frame=2).dtype
```
samuelgarcia: why **end_frame=2** ? **end_frame=1** should be enought ? for security 2 samples are better :)

spikeextractors/cacherecordingextractor.py
alejoe91: ok! I'll remove that. I'll also pass chunk_size

alejoe91:
> Hi Alessio, this is cool.
> Not happy that dtype can be changed, it is dangerous.
> 
> By the way. I always wanted that dtype would be part of the extarctor API like recording.get_dtype()
> Could we put the trick at class level to get the dtype ?

@colehurwitz I think that was the idea right? Is that implemented? In not we should add `get_dtype` in the base RecordingExtractor

colehurwitz:
> > Hi Alessio, this is cool.
> > Not happy that dtype can be changed, it is dangerous.
> > By the way. I always wanted that dtype would be part of the extarctor API like recording.get_dtype()
> > Could we put the trick at class level to get the dtype ?
> 
> @colehurwitz I think that was the idea right? Is that implemented? In not we should add `get_dtype` in the base RecordingExtractor

I don't think we ever talked about doing this, but would we just add, get_dtype in the same way that we added sampling frequency?

alejoe91:
added `get_filename` and `save_fo_file` + tests.

Included https://github.com/SpikeInterface/spikeextractors/pull/265/files

alejoe91:
> > > Hi Alessio, this is cool.
> > > Not happy that dtype can be changed, it is dangerous.
> > > By the way. I always wanted that dtype would be part of the extarctor API like recording.get_dtype()
> > > Could we put the trick at class level to get the dtype ?
> > 
> > 
> > @colehurwitz I think that was the idea right? Is that implemented? In not we should add `get_dtype` in the base RecordingExtractor
> 
> I don't think we ever talked about doing this, but would we just add get_dtype in the same way that we added sampling frequency?

Yes exactly! :) we could also have in the base recording: 
```
get_dtype(self):
    return self.get_traces(start_frame=0, end_frame=1).dtype
```

alejoe91:
this is ready to merge

PR #270, opened 2019-09-30
New recording extractor for nixio (nixpy)

achilleas-k:
This PR adds the NIXIORecordingExtractor for writing to and reading from NIX files using [NIXPy](https://github.com/G-Node/nixpy).  Initial discussion about including NIX can be found in [this issue](https://github.com/G-Node/nixpy/issues/413).

This is an initial attempt based on the contribution guide, the API docs, and the test (adapted from the other RecordingExtractor tests as discussed in the issue linked above).  If it looks like the writer or reader are doing anything out of the ordinary of unexpected, let me know, since I'm not very familiar with SpikeInterface in general.

### Some notes on the implementation
- The `get_traces()` method is a little strange looking and inefficient.  The indexing in NIXPy isn't very flexible in the latest stable release (1.4.9).  The current beta (1.5.0b3) allows fancy indexing like in NumPy and h5py but I didn't want to make the extractor depend on a beta release.  We can clean it up when 1.5.0 gets a stable release.
- At the top level of a NIX file there must always be one or more Block objects.  I used the provided file name to name the Block.  If a file is renamed this wont change the Block name, but the name is inconsequential.  The name can be used to access the Block and it appears when printing but beyond that it has no function.  If there's a better name for it, or if there is some convention in the library, I'd prefer to use that.
- Traces are stored in a DataArray directly as received by the write function.  The DataArray gets one SetDimension, which represents the first dimension (the channels), and a SampledDimension, which represents the second dimension of the data (time).  

### Questions
- The time dimension's units are set as seconds.  Is this correct?  Is it implied that the sampling frequency is in Hz?
- Should the unit of the data be set to volts?  Leaving it undefined (as it is now) wont cause any issues, but for consistency and completeness, it might be better to set it if this is the implied unit of the traces.

---

If this looks good, I'll take a look at writing a SortingExtractor for NIX as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ services:
 python:
 - '3.6'
 install:
-- pip install mountainlab_pytools pynwb pyopenephys kbucket h5py MEArec exdir ruamel.yaml
+- pip install mountainlab_pytools pynwb pyopenephys kbucket h5py MEArec exdir ruamel.yaml nixio
 - pip install .
 - pip install pytest
 script: pytest
--- a/spikeextractors/extractorlist.py
+++ b/spikeextractors/extractorlist.py
@@ -18,6 +18,7 @@ from .extractors.spikeglxrecordingextractor.spikeglxrecordingextractor import Sp
 from .extractors.tridescloussortingextractor.tridescloussortingextractor import TridesclousSortingExtractor
 from .extractors.npzsortingextractor.npzsortingextractor import NpzSortingExtractor
 from .extractors.mcsh5recordingextractor.mcsh5recordingextractor import MCSH5RecordingExtractor
+from .extractors.nixiorecordingextractor.nixiorecordingextractor import NIXIORecordingExtractor
 
 
 recording_extractor_full_list = [
@@ -34,7 +35,8 @@ recording_extractor_full_list = [
     SpikeGLXRecordingExtractor,
     PhyRecordingExtractor,
     MaxOneRecordingExtractor,
-    MCSH5RecordingExtractor
+    MCSH5RecordingExtractor,
+    NIXIORecordingExtractor,
 ]
 
 recording_extractor_dict = {recording_class.extractor_name: recording_class for recording_class in recording_extractor_full_list}
--- a/spikeextractors/extractors/nixiorecordingextractor/__init__.py
+++ b/spikeextractors/extractors/nixiorecordingextractor/__init__.py
@@ -1,0 +1,1 @@
+from .nixiorecordingextractor import NIXIORecordingExtractor
--- a/spikeextractors/extractors/nixiorecordingextractor/nixiorecordingextractor.py
+++ b/spikeextractors/extractors/nixiorecordingextractor/nixiorecordingextractor.py
@@ -1,4 +1,4 @@
-import numpy as np
+import os
 try:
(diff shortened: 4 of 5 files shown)


alejoe91:
Hi @achilleas-k 

Thank you for this! I'll first reply to the questions and then have a closer look at the code.

- the time dimension is in samples/frames. So recording and sorting extractor should implement the get_sampling_frequency method (which returns the fs in Hz) to converto to time. 

- the data is in general uV, but we plan to use the `gain` property (see [here](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/recordingextractor.py#L317)) to be able allow for fast conversion. If the gain is set to 1, then the data are assumed to be in uV. We tried to avoid to use quantities as it requires a lot of extra coding and it can make things harder.

I'll get back with more comments tomorrow (if I have any)

Cheers
Alessio

achilleas-k:
Thanks for the response.  My concern is about saving to the NIX file as much information as is available.  In terms of time, in NIX we use "sampling interval" rather than frequency for regularly sampled signals, so you'll notice the write function stores `1./frequency` and sets the unit of this dimension to `s`.  Based on your answer this should be correct then.

NIX doesn't use quantities for anything complicated (it supports some index scaling between arrays and tags), but it's always good if a file is annotated with this information.

Thanks again.

colehurwitz:
I think it looks great! You can remove the docstrings from the abstract methods as they are documented in the RecordingExtractor by the way. Looks ready to merge though.

colehurwitz:
Ready to merge? I will merge it if you are happy with your current implementation (we are happy as it passes the tests).

colehurwitz:
Thanks for this great contribution!

achilleas-k:
If your'e happy then it's all good!  Merge away.

colehurwitz:
Do you have any comments about the Extractor building process? Was it straightforward? Could we improve our documentation for it?

achilleas-k:
> Do you have any comments about the Extractor building process? Was it straightforward? Could we improve our documentation for it?

Can't say that I have any comments.  It was all pretty straightforward.  The questions and uncertainties I mentioned in the original comment were all NIX specific, so I can't imagine how the docs might cover that.  If I think of anything, or if I come across anything with the SpikeExtractor I'll let you know.

colehurwitz:
I forgot to say, it is very helpful to load the channel locations in during the initialization of the RecordingExtractor (if that information is available).

For example: https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/biocamrecordingextractor/biocamrecordingextractor.py

achilleas-k:
I suppose this is the kind of thing that could be mentioned in the contrib guide (unless it's there and I missed it).  I don't store any properties in NIX as it is now, but it could be added easily.  I'll consult the other writers.  Thanks.

colehurwitz:
I will add this to the docs now! Locations are important for a lot of spike sorters so although it is optional, it would be nice to have. Locations can always be loaded in later manually or with a probe file.

samuelgarcia:
Hi Achileas.
Nice to see you here.

I plan this autumn to make a wrapper on top on neo for all neo rawio.
SO we could get for free all neo reader.
But this will not include writters so this solution is more evolved.

PR #273, opened 2019-10-02
[nixio] Save and load properties

achilleas-k:
This PR adds support for saving channel properties to NIX files and loading them back on read.

The save function creates a metadata section to store channel properties from a RecordingExtractor and attaches the metadata Section to the traces DataArray.
The writer checks the file format version of the NIX library since it will soon change (with the release of 1.5) and the creation of properties will be simplified.

On `__init__` of the RecordingExtractor, the properties are loaded (handled by the `_load_properties()` method).

### DIFF
--- a/spikeextractors/extractors/nixiorecordingextractor/nixiorecordingextractor.py
+++ b/spikeextractors/extractors/nixiorecordingextractor/nixiorecordingextractor.py
@@ -30,6 +30,7 @@ class NIXIORecordingExtractor(RecordingExtractor):
             raise ImportError(missing_nixio_msg)
         RecordingExtractor.__init__(self)
         self._file = nix.File.open(file_path, nix.FileMode.ReadOnly)
+        self._load_properties()
 
     def __del__(self):
         self._file.close()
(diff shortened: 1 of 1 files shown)


achilleas-k:
Working on the SpikeExtractor as well.  Should be ready soon.

colehurwitz:
Awesome! This is a very nice addition to the extractor.

colehurwitz:
One of the booleans we have in the RecordingExtractor is `has_default_locations` which allows us to inform downstream applications if they can expect locations from an extractor (without initializing). I am guessing that there is no way to know this apriori for the NixIORecordingExtractor?

achilleas-k:
> I am guessing that there is no way to know this apriori for the NIXIORecordingExtractor?

I guess there isn't, no.  Is this a useful flag to store if the incoming data from a recording has it set?  Should the writer save the flag and load it back if it's set in the data being stored?

colehurwitz:
I see your point. That's true. Currently, the flag exists as a matter of convenience for a GUI that we develop. I could see it being removed eventually though. I wouldn't worry about it right now.

colehurwitz:
Are you ok to merge this?

achilleas-k:
If you have no further comments or requirements, yeah go ahead.  I'll add the SpikeExtractor in another PR.

colehurwitz:
Sounds great! By the way, the two extractors are RecordingExtractors and SortingExtractors just to clarify the names.

achilleas-k:
My bad.  Misnamed it twice in the comments here, didn't I.  Don't worry, I'm using the correct terminology in the branch https://github.com/G-Node/spikeextractors/tree/nixio-sorting-extractor :)

colehurwitz:
No worries! I was just making sure :)

PR #274, opened 2019-10-03
Updated MCSH5RecordingExtranctor to support changing stream

Shawn-Guo-CN:
Previouly, `MCSH5RecordingExtractor` only supports loading data from 'stream0'. The following changes are made to support changing stream:

1. added 3 new methods, `set_stream_id()`, `get_stream_id()` and `get_available_stream_ids()`;
2. added an argument 'stream_id' to the function `openMCSH5File()` to specify the group path.

@colehurwitz Please test on Linux.


colehurwitz:
Merging

PR #282, opened 2019-10-24
Fix NWB write functions

alejoe91:


alejoe91:
Should we add nwbfile_kwargs is None also to write_recordings?

colehurwitz:
Potentially. Let’s wait on the NWB people responses

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -74,15 +75,31 @@ def __init__(self, file_path, acquisition_name=None):
...
 
-        nwbfile = NWBFile(
-            session_description='',
-            identifier='',
-            session_start_time=datetime.now(),
-        )
+        if os.path.exists(save_path):
+            os.remove(save_path)
+
+        if 'session_description' not in nwbfile_kwargs.keys():
```
bendichter:
```suggestion
        if 'session_description' not in nwbfile_kwargs:
```

bendichter:
This looks good to me. The original idea behind having nwbfile_kwargs be input an whole dict was to allow for kwargs for different neurodata_types, similar to the way seaborn passes args [here](http://seaborn.pydata.org/generated/seaborn.jointplot.html#seaborn.jointplot). I think this is approach is fine though

alejoe91:
> This looks good to me. The original idea behind having nwbfile_kwargs be input an whole dict was to allow for kwargs for different neurodata_types, similar to the way seaborn passes args [here](http://seaborn.pydata.org/generated/seaborn.jointplot.html#seaborn.jointplot). I think this is approach is fine though

Also in seaborn the input is **kwargs so it can be a dictionary or any set of extra keywords. I found a bug in setting the sampling frequency by adding tests (it was multiplying by 1000, when both are in Hz).

Also the SortingExtractor is not implemented. Can you help us including it? See https://github.com/SpikeInterface/spikeextractors/issues/284

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: :+1:

Issue #281, opened 2019-10-24
NwbExtractor fails as session and identifier are not set


alejoe91:
In `write_recording` and `write_sorting`, `session_description` and `identifier` are set to `''`. However, this results in an Error when attempting to open the file with pynwb or when writing sorting data on a file created by `write_recording`.

Suggested fix: 
- add `session_description` and `identifier` as optional args in the write functions
- if None (by default) use `no description` and a random identifier.

@bendichter @luiztauffer does this sound reasonable?

alejoe91:
Also this line: https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L183

Should be:
spkt = sorting.get_unit_spike_train(unit_id=id) / fs

Making a PR now

alejoe91:
https://github.com/SpikeInterface/spikeextractors/pull/282

seancav1967:
Please remove all bugs

bendichter:
@alejoe91 yes, sounds great

alejoe91:
https://github.com/SpikeInterface/spikeextractors/pull/282

PR #285, opened 2019-10-26
Add NWB support

### LINKED ISSUE #284: NwbSortingExtractor is not implemented
@bendichter @luiztauffer 
Can you help me out with this? How do you read units with pynwb?

When I try to access nwbfile.units I get 'Not a dataset'

bendichter:
Sorry for the hold-up. Now that SfN is over we can get back to this. We still need to run a few tests, but here is where we are right now.

fix #284 #281 #282


bendichter:
@alejoe91 do you have any round trip tests we could use to validate this?

alejoe91:
@bendichter I added a simple test that writes a recording and sorting extractor simulated in spikeextractors and then loads it back using the extractors. 
In the near future (@samuelgarcia) we'll have a GIN repo with sample files for each extractor.

Any idea why the test is failing?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -155,23 +405,463 @@ class NwbSortingExtractor(se.SortingExtractor):
...
+            formats as specified by the user.
+        '''
+        # todo
+        return NotImplementedError()
+        assert HAVE_NWB, "To use the Nwb extractors, install pynwb: \n\n pip install pynwb\n\n"
+        if not isinstance(unit_id, int):
+            raise ValueError("'unit_id' must be an integer")
+        existing_ids = self.get_unit_ids()
+        if not isinstance(feature_name, str):
+            raise Exception("'feature_name' must be a string")
```
yarikoptic: TypeError would be a better exception here

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -155,23 +405,463 @@ class NwbSortingExtractor(se.SortingExtractor):
...
+            The unit id for which the features will be set
+        feature_name: str
+            The name of the feature to be stored
+        value
+            The data associated with the given feature name. Could be many
+            formats as specified by the user.
+        '''
+        # todo
+        return NotImplementedError()
+        assert HAVE_NWB, "To use the Nwb extractors, install pynwb: \n\n pip install pynwb\n\n"
```
yarikoptic: code duplication is evil.  I would have created a helper e.g. `check_pynwb` which would have this specific line, avoiding duplication of the message, making it easier to tune (e.g. why not `conda install` if installed via conda etc)

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -155,23 +405,463 @@ class NwbSortingExtractor(se.SortingExtractor):
...
+        if not isinstance(unit_id, int):
+            raise ValueError("'unit_id' must be an integer")
```
yarikoptic: `TypeError` would be a better one here

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -155,23 +405,463 @@ class NwbSortingExtractor(se.SortingExtractor):
...
+
+        with NWBHDF5IO(self._path, 'r') as io:
+            nwbfile = io.read()
+            units = nwbfile.units
+            # chosen unit and interval
+            feat_vals = np.array(units[full_feat_name][units.id.index(unit_id)][:])
+            times = units['spike_times'][units.id.index(unit_id)][:]
+            # spike times are measured in samples
+            frames = self.time_to_frame(times)
+            mask = (frames > start_frame) & (frames < end_frame)
```
yarikoptic:
```suggestion
            mask = (frames >= start_frame) & (frames < end_frame)
```
since doc said "inclusive"

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -155,23 +405,463 @@ class NwbSortingExtractor(se.SortingExtractor):
...
+        with NWBHDF5IO(self._path, 'r+') as io:
+            nwbfile = io.read()
+            nwbfile.add_epoch(start_time=self.frame_to_time(start_frame),
+                              stop_time=self.frame_to_time(end_frame),
+                              tags=epoch_name)
+            io.write(nwbfile)
+
+    def remove_epoch(self, epoch_name=None):
+        '''NWB files do not allow removing epochs.'''
+        print(self.remove_epoch.__doc__)
```
yarikoptic: this common pattern begs to be replaced with some `@justprintthedoc` (or alike) decorator and a pure `pass` in the body.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -155,23 +405,463 @@ class NwbSortingExtractor(se.SortingExtractor):
...
+
+        fs = self.get_sampling_frequency()
+        epoch_info = {}
+        with NWBHDF5IO(self._path, 'r') as io:
+            nwbfile = io.read()
+            flat_list = [item for sublist in nwbfile.epochs['tags'][:] for item in sublist]
+            for i, tag in enumerate(flat_list):
+                if tag == epoch_name:
+                    epoch_info['start_frame'] = int(nwbfile.epochs['start_time'][i] * fs)
+                    epoch_info['end_frame'] = int(nwbfile.epochs['stop_time'][i] * fs)
```
yarikoptic: I wonder if it wouldn't be preferable to use `floor` and `ceil` used instead of pure `int` in both cases here

bendichter:
CI is now passing. We will have some work to do providing read/write access to all the different metadata fields of the file, but it's coming along and currently useable at least on a coarse level

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -1,78 +1,296 @@
...
+def set_dynamic_table_property(dynamic_table, row_ids, property_name, values, default_value=np.nan,
+                               description='no description'):
+    check_nwb_install()
+    if not isinstance(row_ids, list) or not all(isinstance(x, int) for x in row_ids):
+        raise ValueError("'ids' must be an integer or a list of integers")
+    ids = list(dynamic_table.id[:])
+    if any([i not in ids for i in row_ids]):
+        raise ValueError("'ids' contains values outside the range of existing ids")
+    if not isinstance(property_name, str):
+        raise Exception("'property_name' must be a string")
```
yarikoptic: TypeError here and below. Raising pure Exceptions is not recommended since complicates handling etc

bendichter:
@yarikoptic always great to have experienced eyes on code QC. Thanks for looking through it

alejoe91:
@bendichter @yarikoptic let me know when you are ready to merge! Looks like very good progress to me!

bendichter:
Thanks @alejoe91, I agree this is coming along. @luiztauffer has been helping to build this and test all the features are working as expected. Luiz, what is the status? Do you think this is ready for review?

luiztauffer:
few things remaining, that were not covered by the CI tests, I expect to finish it in the coming days.

tests/test_extractors.py
```diff
@@ -236,7 +236,9 @@ def test_nwb_extractor(self):
         # create new
         path2 = self.test_dir + '/firings_true.nwb'
         se.NwbSortingExtractor.write_sorting(self.SX, path2, session_description='second',
-                                                 identifier='19475')
+                                             identifier='19475')
+        SX_nwb = se.NwbSortingExtractor(path1)
```
alejoe91: @luiztauffer this should be `path2`, but locally it then fails because `No acquisitions found in the .nwb file.`
You probably need to add a dummy acquisition if only spiketrains are saved

alejoe91:
@luiztauffer @bendichter the current h5 slicer fails if the channels are not ordered (this might happen after a probe file is loader)

Check out the MEArec solution: https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/mearecextractors/mearecextractors.py#L81-L89

Basically I reorder channels before retrieval and then order them back to the original channel map

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -1,78 +1,296 @@
...
+    if not isinstance(property_name, str):
+        raise TypeError("'property_name' must be a string")
+    if len(row_ids) != len(values):
+        raise ValueError("'ids' and 'values' should be lists of same size")
 
-    def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
-        return self._other.get_traces(channel_ids=channel_ids, start_frame=start_frame, end_frame=end_frame)
+    if property_name in dynamic_table:
+        for (row_id, value) in zip(row_ids, values):
+            dynamic_table[property_name][ids.index(row_id)] = value
```
alejoe91: When I try: `recording.set_channel_property(1, 'group', 0)` I get the following error at this line:
`TypeError: 'VectorData' object does not support item assignment`

luiztauffer: thanks for pointing that out! it should be fixed in the coming PR update

luiztauffer:
> @luiztauffer @bendichter the current h5 slicer fails if the channels are not ordered (this might happen after a probe file is loader)
> 
> Check out the MEArec solution: [/spikeextractors/extractors/mearecextractors/mearecextractors.py@`master`#L81-L89](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/mearecextractors/mearecextractors.py#L81-L89) [](/SpikeInterface/spikeextractors/blob/HEAD@{2019-11-04T11:16:38Z}/spikeextractors/extractors/mearecextractors/mearecextractors.py#L81-L89)
> 
> Basically I reorder channels before retrieval and then order them back to the original channel map

Oh, I see. Thanks for finding that, I'll fix it

colehurwitz:
Sorry, I just got a chance to look at this PR, but it seems like you are overwriting some functions like `set_unit_spike_features`.

All non-abstract functions (except for write_recording/write_sorting) are generic for any spikeextractor. So functions like `set_unit_spike_features` are implemented only at a parent class level and then all extractors can use them. Unless I am misunderstanding your intention, I would not recommend overwriting anything but the required functions and the write functions.

alejoe91:
> Sorry, I just got a chance to look at this PR, but it seems like you are overwriting some functions like `set_unit_spike_features`.
> 
> All non-abstract functions (except for write_recording/write_sorting) are generic for any spikeextractor. So functions like `set_unit_spike_features` are implemented only at a parent class level and then all extractors can use them. Unless I am misunderstanding your intention, I would not recommend overwriting anything but the required functions and the write functions.

Hi @colehurwitz I think that the idea is to dynamically update the underlying nwb file

bendichter:
@alejoe91 and @colehurwitz yes, that is the intention

bendichter:
oh, is the idea to build up the spikeextractor in memory and then write it all at once?

colehurwitz:
Ah, I see. So the dictionary won't be used at all, but rather it will always be using the underlying nwb file. This is cool. Ignore me!

tests/test_extractors.py
luiztauffer: you're right! we're working on it

bendichter:
We could have done it by building up the spikeextractor in memory and then write it all at once. That is closer to how pynwb works as well. That wasn't clear to me. If you want people to develop extensions in that way, I suggest you add a guide in the README that tells them precisely what methods you need to write to build a new extractor

alejoe91:
> oh, is the idea to build up the spikeextractor in memory and then write it all at once?

That could also be a viable solution. I think it's better actually. Sometimes one might try something out and accidentally overwrite some fields in the file..If the user is required to save then he/she will make sure that everything looks good. Do you agree @bendichter ?

bendichter:
I could go either way.

alejoe91:
@colehurwitz what do you think?

alejoe91:
> We could have done it by building up the spikeextractor in memory and then write it all at once. That is closer to how pynwb works as well. That wasn't clear to me. If you want people to develop extensions in that way, I suggest you add a guide in the README that tells them precisely what methods you need to write to build a new extractor

I think it says here that all you need to do is write the 4 abstract methods (+ the write_recording optionally): https://spikeinterface.readthedocs.io/en/latest/recording_extractor.html

bendichter:
ah ok I missed that

colehurwitz:
I think it is probably easier if all the regular functions are used and then it is written it all at once. 

For our documentation, I think we can be more clear about which functions are implemented in the parent class and should not be implemented in the children classes. I can change that to avoid confusion.

bendichter:
here are the pros and cons for doing it in memory as I see it

**pros:**
* uniform behavior with other extractors
* less nwb-specific code to maintain

**cons:**
* could take up much more memory in some cases, particularly if you have unit_spike_features
* if you want to just change one little thing, you may need to rewrite the entire file. We might be able to be clever about this though.

I'm pretty split on this I guess. If you think going in-memory is better we can go that way.

alejoe91:
What about having the possibility to do both? for example we could have an `in_place` argument in the init.

colehurwitz:
I think it is better to do in-memory since it is more inline with what the rest of the spikeextractors are doing... we can always add some way of storing temporary files for these dictionaries later or something if we are concerned about memory. I don't think memory will be an issue for most features (except maybe waveform).

colehurwitz:
@bendichter @luiztauffer can I close this one?

PR #293, opened 2019-11-04
unique id (from counter)

colehurwitz:
I wanted a unique ID for each extractor that was created in a session. This is useful for naming temporary files, etc., when working with many Recordings and Sortings. What do you think @samuelgarcia ?

### DIFF
--- a/spikeextractors/recordingextractor.py
+++ b/spikeextractors/recordingextractor.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 import numpy as np
 import copy
+import random
 from .extraction_tools import load_probe_file, save_to_probe_file, write_to_binary_dat_format, get_sub_extractors_by_property
 
 class RecordingExtractor(ABC):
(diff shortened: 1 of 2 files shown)


alejoe91:
Are you sure that this would be updated by initializing multiple extractors?

colehurwitz:
I can add a test for it. I tried it with multiple extractors and multiple subextractors and it worked.

alejoe91:
Let's see what Sam says :)

colehurwitz:
Sure, i wanted a unique identifier so I could save out temporary files like this:

`file_name = 'recording_' + str(recording.id)`

colehurwitz:
Whoops, didn't realize that referencing this PR elsewhere would do this... this PR has nothing to do with NEO!

samuelgarcia:
Hi Cole.
I am OK with this if you need this for your workflow.
The only drwback I can see is that if you run in parralel in 2 python interpreters then this ID will unique inside each interpreters but not globally unique and so if you use this ID to generate temporary files you could have colision. No ?

colehurwitz:
Is there a way to make something like this work in different interpreters or is that impossible?

samuelgarcia:
Maybe with random at unit and then counter.

colehurwitz:
I think actually just giving a random number upon init should work. I removed the counter because it doesn't really add much when making temporary files.

PR #295, opened 2019-11-04
rEmoved inplace probe loading

colehurwitz:
@alejoe91

### DIFF
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -74,7 +74,6 @@ def load_probe_file(recording, probe_file, channel_map=None, channel_groups=None
         The extractor containing all of the probe information.
     '''
     from .subrecordingextractor import SubRecordingExtractor
-    from .subsortingextractor import SubSortingExtractor
     probe_file = Path(probe_file)
     if probe_file.suffix == '.prb':
         probe_dict = read_python(probe_file)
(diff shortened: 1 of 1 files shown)


colehurwitz:
This seems like it will still work and passes all tests. Let me know what you think please.

spikeextractors/extraction_tools.py
```diff
@@ -86,14 +85,10 @@ def load_probe_file(recording, probe_file, channel_map=None, channel_groups=None
...
-
-            if list(ordered_channels) == recording.get_channel_ids():
-                subrecording = recording
-            else:
-                if not np.all([chan in recording.get_channel_ids() for chan in ordered_channels]) and verbose:
-                    print('Some channel in PRB file are not in original recording')
-                present_ordered_channels = [chan for chan in ordered_channels if chan in recording.get_channel_ids()]
-                subrecording = SubRecordingExtractor(recording, channel_ids=present_ordered_channels)
+            if not np.all([chan in recording.get_channel_ids() for chan in ordered_channels]) and verbose:
+                print('Some channel in PRB file are not in original recording')
```
samuelgarcia: I would raise an error instead no ?

alejoe91: No this is perfectly fine. Some channels could be ref and users might want to exclude them from the probe file..I think we discussed about it already!

alejoe91:
@colehurwitz looks good to me

samuelgarcia:
@colehurwitz @alejoe91 
I have just realise that this PR have broken a important feature for me.

In some sorters (KS, KS2, klusta, TDC) to avoid a heavy copy each time I tested in sorter if the input was an BinDatRecordingExtractor with **isinstance(recording, se.BinDatRecordingExtractor)**.
When it was the case then no local copy and instead the filename was given to the sorter.
It was very convinient in ComparisonStudy because the input is always BinDatRecordingExtractor and so no copy was perform at all.

Now with this **not inplace load probe**, the recording is not anymore a BinDatRecordingExtractor but indeed a SubRecordingExtractor.
So the feature is broken for me and it was quite important.
How could we find a solution for this.

Copy everything each time for each sorter is so painfull we I work on benchmark. How could we fix that ?

alejoe91:
Mmm good point @samuelgarcia...I don't know what the best would be. Maybe for subrecording extractor to store the type of the parent extractor? Unfortunately I think that loading the probe file must return a subrecording, otherwise reordering/skipping channels gets complicated

samuelgarcia:
Totally agree the load_probe must not be in place and have to return subrecording.
But the need of not copy everything is really important.

I think a quick and dirty fix could be to make a load_probe_in_place that would be used only in the context of ComparisonStudy.
In comparison study, the recording is copied (and shared) inside "raw_files" folder. In that case the channel map is complete and the PRB is one group always. It is a simple case.

So in that specific case we could have a load_probe_file_inplace that would be call by ComparisonStudy when loading the PRB. This load_probe_file_inplace could be in studytools and not in spikeextractor to avoid mistake.

What do you think ?

PR #297, opened 2019-11-12
add shybrid compatibility to spikeinterface

jwouters91:
- added shybrid recording and sorting extractor
- also implemented the static write methods for both extractors

### DIFF
--- a/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
+++ b/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
@@ -36,6 +36,8 @@ class BinDatRecordingExtractor(RecordingExtractor):
         self._geom = geom
 
         if recording_channels is not None:
+            assert len(recording_channels) == self._timeseries.shape[0], \
+               'Provided recording channels have the wrong length'
             self._channels = recording_channels
         else:
             self._channels = list(range(self._timeseries.shape[0]))
--- a/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
+++ b/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
@@ -36,15 +36,13 @@ class BinDatRecordingExtractor(RecordingExtractor):
         self._geom = geom
 
         if recording_channels is not None:
-            assert len(recording_channels) == self._timeseries.shape[0], \
-                'Provided recording channels have the wrong length'
             self._channels = recording_channels
         else:
             self._channels = list(range(self._timeseries.shape[0]))
 
         if geom is not None:
-            for m in range(self._timeseries.shape[0]):
-                self.set_channel_property(m, 'location', self._geom[m, :])
+            for idx, channel in enumerate(self._channels):
+                self.set_channel_property(channel, 'location', self._geom[idx, :])
 
     def get_channel_ids(self):
         return self._channels
--- a/spikeextractors/extractors/shybridextractors/shybridextractors.py
+++ b/spikeextractors/extractors/shybridextractors/shybridextractors.py
@@ -4,7 +4,7 @@ import numpy as np
 
 from spikeextractors import SortingExtractor
 from spikeextractors.extractors.bindatrecordingextractor import BinDatRecordingExtractor
-from spikeextractors.extraction_tools import save_to_probe_file
+from spikeextractors.extraction_tools import save_to_probe_file, load_probe_file
 
 try:
     import hybridizer.io as sbio
@@ -44,6 +44,8 @@ class SHYBRIDRecordingExtractor(BinDatRecordingExtractor):
                                           nb_channels,
                                           params['dtype'],
                                           time_axis=time_axis)
+
+        self = load_probe_file(self, params['probe'])
 
     @staticmethod
     def write_recording(recording, save_path, initial_sorting_fn, dtype='float32'):
--- a/spikeextractors/extractors/shybridextractors/shybridextractors.py
+++ b/spikeextractors/extractors/shybridextractors/shybridextractors.py
@@ -82,17 +82,11 @@ class SHYBRIDRecordingExtractor(BinDatRecordingExtractor):
         save_to_probe_file(recording, probe_fn)
 
         # create parameters file
-        source_dir, _ = os.path.split(__file__)
-        parameters_template_fn = os.path.join(source_dir, 'parameters_template')
-
-        with open(parameters_template_fn, 'r') as fp:
-            parameters_template = fp.read()
-
-        parameters = parameters_template.format(initial_sorting_fn=initial_sorting_fn,
-                                                data_type=dtype,
-                                                sampling_frequency=str(recording.get_sampling_frequency()),
-                                                byte_ordering='F',
-                                                probe_fn=probe_fn)
+        parameters = params_template.format(initial_sorting_fn=initial_sorting_fn,
+                                            data_type=dtype,
+                                            sampling_frequency=str(recording.get_sampling_frequency()),
+                                            byte_ordering='F',
+                                            probe_fn=probe_fn)
 
         # write parameters file
         parameters_fn = os.path.join(save_path, PARAMETERS_NAME)
@@ -160,3 +154,13 @@ class GeometryNotLoadedError(Exception):
     """ Raised when the recording extractor has no associated channel locations
     """
     pass
+
+params_template = \
+"""clusters:
+  csv: {initial_sorting_fn}
+data:
+  dtype: {data_type}
+  fs: {sampling_frequency}
+  order: {byte_ordering}
+  probe: {probe_fn}
+"""
--- a/spikeextractors/extractors/shybridextractors/shybridextractors.py
+++ b/spikeextractors/extractors/shybridextractors/shybridextractors.py
@@ -1,6 +1,9 @@
+import os
+
+import numpy as np
+
 from spikeextractors import RecordingExtractor
 from spikeextractors import SortingExtractor
-
 from spikeextractors.extractors.bindatrecordingextractor import BinDatRecordingExtractor
 
 try:
@@ -24,7 +27,7 @@ class SHYBRIDRecordingExtractor(RecordingExtractor):
         self._convert_shybrid_probe()
 
         # translate the byte ordering
-        # TODO still ambiguous, shybrid assumes time_axis=1
+        # TODO still ambiguous, shybrid should assume time_axis=1, since spike interface makes an assumption on the byte ordering
         byte_order = self._params['order']
         if byte_order == 'C':
             time_axis = 1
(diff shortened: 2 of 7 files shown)


alejoe91:
@jwouters91 can you merge with the latest master to resolve the conflicts?

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
```diff
@@ -36,15 +36,13 @@ def __init__(self, file_path, sampling_frequency, numchan, dtype, recording_chan
         self._geom = geom
 
         if recording_channels is not None:
-            assert len(recording_channels) == self._timeseries.shape[0], \
```
alejoe91: this assertion should be satisfied in my opinion. Is there a reason why you removed it?

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
```diff
@@ -36,15 +36,13 @@ def __init__(self, file_path, sampling_frequency, numchan, dtype, recording_chan
...
-                'Provided recording channels have the wrong length'
             self._channels = recording_channels
         else:
             self._channels = list(range(self._timeseries.shape[0]))
 
         if geom is not None:
-            for m in range(self._timeseries.shape[0]):
-                self.set_channel_property(m, 'location', self._geom[m, :])
+            for idx, channel in enumerate(self._channels):
+                self.set_channel_property(channel, 'location', self._geom[idx, :])
```
alejoe91: thanks :)

spikeextractors/extractors/shybridextractors/shybridextractors.py
```diff
@@ -0,0 +1,174 @@
...
+except ImportError:
+    HAVE_SBEX = False
+
+class SHYBRIDRecordingExtractor(RecordingExtractor):
+    extractor_name = 'SHYBRIDRecordingExtractor'
+    installed = HAVE_SBEX
+    extractor_gui_params = [
+        {'name': 'recording_fn', 'type': 'file', 'title': "Full path to hybrid recording (.bin)"},
+    ]
+
```
alejoe91: can you add an installation_mesg?

spikeextractors/extractors/shybridextractors/shybridextractors.py
```diff
@@ -0,0 +1,174 @@
...
+from spikeextractors.extraction_tools import save_to_probe_file
+
+try:
+    import hybridizer.io as sbio
+    import hybridizer.probes as sbprb
+    HAVE_SBEX = True
+except ImportError:
+    HAVE_SBEX = False
+
+class SHYBRIDRecordingExtractor(RecordingExtractor):
```
alejoe91: you can also derive the class directly from `BinDatRecordingExtractor` so you don't need to reimplement the abstract methods

spikeextractors/extractors/shybridextractors/shybridextractors.py
```diff
@@ -0,0 +1,174 @@
...
+    extractor_name = 'SHYBRIDRecordingExtractor'
+    installed = HAVE_SBEX
+    extractor_gui_params = [
+        {'name': 'recording_fn', 'type': 'file', 'title': "Full path to hybrid recording (.bin)"},
+    ]
+
+    def __init__(self, recording_fn):
```
alejoe91: the argument should be `folder_path` or `file_path`

spikeextractors/extractors/shybridextractors/shybridextractors.py
```diff
@@ -0,0 +1,174 @@
...
+            fp.write(parameters)
+
+
+class SHYBRIDSortingExtractor(SortingExtractor):
+    extractor_name = 'SHYBRIDSortingExtractor'
+    installed = HAVE_SBEX
+    extractor_gui_params = [
+        {'name': 'gt_fn', 'type': 'file', 'title': "Full path to hybrid ground truth labels (.csv)"},
+    ]
+    is_writable = False
```
alejoe91: this should be `True` as you have a write function. Can you add it to the RecordingExtractor as well?

spikeextractors/extractors/shybridextractors/shybridextractors.py
```diff
@@ -0,0 +1,174 @@
...
+        parameters
+        ----------
+        recording: RecordingExtractor
+            The recording extractor to be converted and saved
+        save_path: str
+            Full path to desired target folder
+        initial_sorting_fn: str
+            Full path to the initial sorting csv file (can also be generated
+            using write_sorting static method from the SHYBRIDSortingExtractor)
+        dtype: dtype
```
alejoe91: you can also infer the dtype from the recording. We'll make a method of the base extractor soon (so this is ok)

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
jwouters91: I removed the assertion when I was trying to make the binary data recording extractor support broken channels. In case of broken channels, you want to subsample the number of channels. Because of this, the length of the provided list of channels is not necessarily equal to the true number of channels that are present in the recording. I realize now that it might be better to use a SubRecordingExtractor here.

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
jwouters91: Not sure anymore if this is the best way for supporting broken channels...

spikeextractors/extractors/shybridextractors/shybridextractors.py
jwouters91: good idea!

spikeextractors/extractors/shybridextractors/shybridextractors.py
jwouters91: The installation procedure is still to be improved, let me get back to this point later.

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
alejoe91: Also in the probe file you can skip channels! It will extract a SubRecordingExtractor automatically

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
alejoe91: I think your change is correct here. If one passes the channel_ids then we should loop through it

spikeextractors/extractors/shybridextractors/shybridextractors.py
jwouters91: I don't fully get this comment. Do you want me to take any action regarding this topic right now?

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
alejoe91: @jwouters91 can you add the assertion back?

spikeextractors/extractors/shybridextractors/shybridextractors.py
```diff
@@ -0,0 +1,159 @@
...
+        # translate the byte ordering
+        # TODO still ambiguous, shybrid should assume time_axis=1, since spike interface makes an assumption on the byte ordering
+        byte_order = params['order']
+        if byte_order == 'C':
+            time_axis = 1
+        elif byte_order == 'F':
+            time_axis = 0
+
+        # piggyback on binary data recording extractor
+        BinDatRecordingExtractor.__init__(self,
```
alejoe91: would the broken channels be in the probe file? In that case, you could add this line after the init:

`self = se.load_probe_file(self, probe)`

This will return a SubRecordingExtractor with the channel map and your object will be set to this new extractor.
Locations will be also loaded automatically.
Does it make sense?

spikeextractors/extractors/shybridextractors/shybridextractors.py
alejoe91: you could automatically extract the dtype as follows:
`dtype = recording.get_traces(start_frame=0, end_frame=1).dtype`

I think it would be better. I agree that my previous comment was very confusing!

spikeextractors/extractors/shybridextractors/shybridextractors.py
jwouters91: The probe file indeed contains the information about the broken channels. As per your suggestion (https://github.com/jwouters91/shybrid/issues/1#issuecomment-536643086) I removed loading the probe file completely from the extractor. Now the user will have to explicitly load the probe file (https://github.com/jwouters91/shybrid/blob/f33e4fc732fcb3d6ca76182265c349ff229f4382/examples/spikeinterface_validation_example.py#L22)

Which style do you think is most in line with the framework?

spikeextractors/extractors/shybridextractors/shybridextractors.py
jwouters91: Here the dtype reflects the type that user wants to save the data to, this could be potentially different from the data type that is used internally in spikeinterface.

alejoe91: ah! Ok then leave it there :)

spikeextractors/extractors/shybridextractors/shybridextractors.py
alejoe91: I guess I don't agree with my past self...I think at that time I didn't realize that the probe file is expected to be there in SHYBRID format. I think it would be nice to load it automatically. Sorry for my swinging opinions

alejoe91:
One last thing: can you add a basic test in the `tests/test_extractors.py`? There are plenty of examples there. The idea is to write a recording and a sorting extractors (self._RX and self._SX), load them back, and check that the results are the same

jwouters91:
I assume the build fails, because shybrid is not yet installed in the test env. On my dev environment shybrid tests passed properly.

alejoe91:
You can add in the .travis `pip install https://github.com/jwouters91/shybrid/archive/master.zip`

alejoe91:
Uh sorry we made a modification to the save_to_probe_file in the meanwhile. Now it doesn't need the format arg anymore. Can you merge with master? I think that this might be why it fails

jwouters91:
> Uh sorry we made a modification to the save_to_probe_file in the meanwhile. Now it doesn't need the format arg anymore. Can you merge with master? I think that this might be why it fails

No worries, that's why we need tests :)

jwouters91:
Let me get back later to solving this one, this error is related to the travis env and package installation. Probably only get to look at it on Monday.

jwouters91:
Finally streamlined the install procedure and added the installation message to the extractors. Let me known if you think  any other changes are needed prior to merging.

alejoe91:
looks good to me! I'll ask @colehurwitz to take a look too and then we can merge :)

colehurwitz:
Looks great! I am currently traveling and can’t look super deeply, but I think it looks ready from what I have seen.

alejoe91:
I'll merge this @jwouters91. Thankg again!

jwouters91:
Do you also want me to fork and adapt the docs? Or do you prefer another way of working there?

alejoe91:
That would be nice, thanks! can you make a PR to spikeinterface for it?

Issue #298, opened 2019-11-13
sorting.get_unit_spike_features() does not work as intended with frame ranges


luiztauffer:
I believe `sorting.get_unit_spike_features()` method is not working as intended when the user specifies a frame range.
I think the problem is in this line:
https://github.com/SpikeInterface/spikeextractors/blob/58fafe687ebb828fb465bc7240caf05bbd662d78/spikeextractors/sortingextractor.py#L157

Shouldn't only the indexes of spikes that fall within the defined range be used here instead of the frame range?

colehurwitz:
Yes, I think you are right! This line doesn't make sense since it is trying to index the actual feature. I can fix this.

alejoe91:
@luiztauffer thanks for spotting this and @colehurwitz for fixing :)

colehurwitz:
Fixed! #300

PR #300, opened 2019-11-13
fix index issue spike features

colehurwitz:
#298 (fix in sortingextractor, subsortingextractor, and extra tests).


colehurwitz:
@luiztauffer @alejoe91

PR #305, opened 2019-11-27
added get_dtype to recoridngectractor

alejoe91:


spikeextractors/recordingextractor.py
```diff
@@ -99,6 +99,9 @@ def get_num_channels(self):
         #       'This warning will be removed in future versions of SpikeInterface.')
         return len(self.get_channel_ids())
 
+    def get_dtype(self):
+        return self.get_traces(channel_ids=self.get_channel_ids()[0], start_frame=0, end_frame=1).dtype
+
```
samuelgarcia: I wanted to do that since lot of time.
This can cost a bit because it call get_traces.
Sometime the dtype can be known before the get_tarce.
I would propose 
```python
def get_dtype(self):
   return self._get_dtype()
def _get_dtype(self):
   # This should be overwritten in subclass
   return self.get_traces(channel_ids=self.get_channel_ids()[0], start_frame=0, end_frame=1).dtype


```

So we can overwritte this method in some class to be more efficient.

alejoe91: or we can have get_dtype be overridden directly :)

alejoe91: If extractors are lazy, this just loads a single value (1 channel, 1 timepoint)

samuelgarcia: Yes I known the underscore encourage to overwrite the method I think.
It is more clear.
I won't debate on this :)

alejoe91: @samuelgarcia I think this is ok actually.

Can we merge it?

PR #314, opened 2020-01-17
memory issues spikeglx (fix)

colehurwitz:
Spikeglx was instantiating (but not using) two arrays that were the same size as the data! I disabled those..


colehurwitz:
Not sure why it failed the first time... NWB has some randomness potentially? Weird. @alejoe91

PR #316, opened 2020-01-29
Implement dump and load_extractor_from_json

alejoe91:
Implemented dump for MEArec, BinDat (and Cache), and Mda (recording and sorting)

The idea is to define the `self.kwargs` dictionary at initialization and dump the extractor_name and the params to json (from base extractors). The extractors can be reloaded with the `load_extractor_from_json` function.
Only extractor that have `is_dumpable` True can be dumped and loaded.

Minimal example:
```
import spikeextractors as se
import numpy as np

rec, sort = se.example_datasets.toy_example(duration=10, num_channels=4)
se.NumpyRecordingExtractor.is_dumpable  # no
se.MdaRecordingExtractor.is_dumpable  # yes

# save and load mda
se.MdaRecordingExtractor.write_recording(rec, 'rec_folder')
rec_mda = se.MdaRecordingExtractor('rec_folder')
rec_mda.dump()  # this saves 'spikeinterface_recording.json'

# load from json
rec_mda_loaded = se.load_extractor_from_json('spikeinterface_recording.json')
assert np.array_equal(rec_mda.get_traces(), rec_mda_loaded.get_traces())
```

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -11,44 +11,65 @@ class BaseExtractor:
 
     def __init__(self):
         self._kwargs = {}
-        self._dump_dict = {}
+        self._tmp_folder = None
 
     def make_serialized_dict(self):
         class_name = str(type(self)).replace("<class '", "").replace("'>", '')
         module = class_name.split('.')[0]
         imported_module = importlib.import_module(module)
 
-        dump_dict = {'class': class_name, 'module': module, 'kwargs': self._kwargs,
-                     'version': imported_module.__version__}
+        if self.is_dumpable:
+            if 'CacheRecording' in class_name:
+                print("Warning: dumping a CacheRecordingExtractor. The path to the tmp binary file will be lost in "
+                      "further sessions. To prevent this, use the "
+                      "'CacheRecordingExtractor.save_to_file('path-to-file)' function")
+
+            dump_dict = {'class': class_name, 'module': module, 'kwargs': self._kwargs,
+                         'version': imported_module.__version__, 'dumpable': True}
+
+            if 'Recording' in class_name:
+                if 'group' in self.get_shared_channel_property_names():
+                    groups = self.get_channel_groups()
+                    dump_dict['groups'] = groups
+                if 'location' in self.get_shared_channel_property_names():
+                    locations = self.get_channel_locations()
+                    dump_dict['locations'] = locations
+            elif 'Sorting' in class_name:
+                if 'group' in self.get_shared_unit_property_names():
+                    groups = [self.get_unit_property(u, 'group') for u in self.get_unit_ids()]
+                    dump_dict['groups'] = groups
+        else:
+            dump_dict = {'class': class_name, 'module': module, 'kwargs': {},
+                         'version': imported_module.__version__, 'dumpable': False}
 
         return dump_dict
 
-    def dump(self, output_folder=None, output_filename=None):
+    def dump(self, folder_path=None, file_name=None):
         '''
         Dumps recording extractor to json file.
         The extractor can be re-loaded with spikeextractors.load_extractor_from_json(json_file)
 
         Parameters
         ----------
-        output_folder: str or Path
+        folder_path: str or Path
             Path to output_folder
-        output_filename: str
+        file_name: str
             Filename
         '''
-        if output_folder is None:
-            output_folder = Path(os.getcwd())
-        else:
-            output_folder = Path(output_folder)
-        if output_filename is None:
-            output_filename = 'spikeinterface_recording.json'
-        elif Path(output_filename).suffix == '':
-            output_filename = output_filename + '.json'
-        dump_dict = self.make_serialized_dict()
-        if self.is_dumpable:
-            with open(str(output_folder / output_filename), 'w', encoding='utf8') as f:
+        if self.check_if_dumpable():
+            if folder_path is None:
+                folder_path = Path(os.getcwd())
+            else:
+                folder_path = Path(folder_path)
+            if file_name is None:
+                file_name = 'spikeinterface_recording.json'
+            elif Path(file_name).suffix == '':
+                file_name = file_name + '.json'
+            dump_dict = self.make_serialized_dict()
+            with open(str(folder_path / file_name), 'w', encoding='utf8') as f:
                 json.dump(_check_json(dump_dict), f, indent=4)
         else:
-            raise Exception(f"Object {str(type(self))} is not dumpable to json")
+            raise Exception(f"The extractor is not dumpable to to json")
 
     @staticmethod
     def load_extractor_from_json(json_file):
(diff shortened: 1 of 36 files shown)


alejoe91:
@samuelgarcia what do you think?

samuelgarcia:
Hi Alessio.
This is very cool and fit with what I had in mind.
Maybe it could be safe to be able to select between json and pickle because for JSON the dump can be problematic for instance with np.int16, np.float32 types. See the PR I have done for spieksorters.

For dump we could have the option ```format=None/'json'/'pickle'``` None would be auto select with a preference for json but if it fails then it would be pickle.
What do you think ?

alejoe91:
In your PR you have the _check_json. That should work here too right?

samuelgarcia:
Yes you should include this _check_json but you will discover that there is always a new exception in types not supported by json and the elsif case will grow sometimes.

For the moment, this is OK, we will see in the long run if it is stable enought with all extractor (included neo ones)

spikeextractors/extractors/neuroscopesortingextractor/neuroscopesortingextractor.py
```diff
@@ -41,6 +41,8 @@ def __init__(self, resfile, clufile):
         else:
             self._spiketrains = []
             self._unit_ids = []
+        self._kwargs = {'resfile': resfile, 'clufile': clufile}
```
samuelgarcia: str(resfile)
str(clufile)

spikeextractors/extractors/nixioextractors/nixioextractors.py
```diff
@@ -168,6 +168,7 @@ def __init__(self, file_path):
             sfreq = md["sampling_frequency"]
             self._sampling_frequency = sfreq
         self._load_properties()
+        self._kwargs = {'file_path': file_path}
```
samuelgarcia: And same everywhere it is a file.

spikeextractors/baseextractor.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Returns
+        -------
+        extractor: RecordingExtractor or SortingExtractor
+            The loaded extractor object
+        '''
+        with open(str(json_file), 'r') as f:
+            d = json.load(f)
+            extractor = _load_extractor_from_dict(d)
+        return extractor
```
samuelgarcia: OK to have this class method.
We should also have the same method but with the dict in case of not dumped to a file (with pickle or dask, joblib).

spikeextractors/baseextractor.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _load_extractor_from_dict(dic, check_version=False):
```
samuelgarcia: _load_extractor_from_dict > load_extractor_from_dict

spikeextractors/baseextractor.py
```diff
@@ -0,0 +1,116 @@
...
+    extractor = None
+    kwargs = dic['kwargs']
+    if np.any([isinstance(v, dict) for v in kwargs.values()]):
+        for k in kwargs.keys():
+            if isinstance(kwargs[k], dict):
+                if 'module' in kwargs[k].keys() and 'class' in kwargs[k].keys() and 'version' in kwargs[k].keys():
+                    extractor = _load_extractor_from_dict(kwargs[k])
+                    class_name = dic['class']
+                    cls = _get_class_from_string(class_name)
+                    if not _check_same_version(class_name, dic['version']):
```
samuelgarcia: You are a friend.

spikeextractors/extractors/biocamrecordingextractor/biocamrecordingextractor.py
```diff
@@ -155,23 +160,26 @@ def readHDF5t_100(rf, t0, t1, nch):
...
 
 def readHDF5t_101(rf, t0, t1, nch):
     if t0 <= t1:
-        return rf['3BData/Raw'][nch * t0:nch * t1].reshape((t1-t0, nch), order='C')
+        return rf['3BData/Raw'][nch * t0:nch * t1].reshape((t1 - t0, nch), order='C')
     else:  # Reversed read
         raise Exception('Reading backwards? Not sure about this.')
-        return rf['3BData/Raw'][nch * t1:nch * t0].reshape((t1-t0, nch), order='C')
+        return rf['3BData/Raw'][nch * t1:nch * t0].reshape((t1 - t0, nch), order='C')
+
```
samuelgarcia: Totay exceptional.

spikeextractors/baseextractor.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _load_extractor_from_dict(dic, check_version=False):
+    extractor = None
+    kwargs = dic['kwargs']
+    if np.any([isinstance(v, dict) for v in kwargs.values()]):
+        for k in kwargs.keys():
+            if isinstance(kwargs[k], dict):
+                if 'module' in kwargs[k].keys() and 'class' in kwargs[k].keys() and 'version' in kwargs[k].keys():
+                    extractor = _load_extractor_from_dict(kwargs[k])
```
samuelgarcia: This is really cool, the nested stuff and finally more easy, no ?

samuelgarcia:
Alessio, this is really really cool.

alejoe91:
95% ready! just need to add:
- a custom save/load of groups and locations for recordings and groups for sorting.
- tests

alejoe91:
@samuelgarcia tomorrow I'll serialize spike sorting! Then you can start the dask backend ;)

spikeextractors/baseextractor.py
alejoe91: I added a static method and methods in the extraction tools.

So you can call directly: `se.load_extractor_from_dict(d)`

spikeextractors/baseextractor.py
alejoe91: done

spikeextractors/baseextractor.py
alejoe91: Yes it was fun! Now you have no excuse to implement the dask backend ehehe

spikeextractors/recordingextractor.py
```diff
@@ -252,19 +250,21 @@ def get_channel_locations(self, channel_ids=None):
         for channel_id in channel_ids:
             location = self.get_channel_property(channel_id, 'location')
             locations.append(location)
-        return locations
+        return np.array(locations)
 
-    def set_channel_groups(self, channel_ids, groups):
+    def set_channel_groups(self, groups, channel_ids=None):
```
alejoe91: most of the times you want to set the groups for all channels...

samuelgarcia: very dangerous.
This args invertion will break many codes.
I am ok with this but need to be tested in all package and published notebooks.

alejoe91: I'm doing it :) It was very annoying...of course I want all channels by default!

alejoe91:
merging this!

PR #317, opened 2020-01-29
Set chunk size in MB (500 by default)

alejoe91:


alejoe91:
@samuelgarcia something like this shoulw work :)

alejoe91:
can we merge this?

spikeextractors/extraction_tools.py
```diff
@@ -267,25 +267,54 @@ def write_to_binary_dat_format(recording, save_path, time_axis=0, dtype=None, ch
...
     '''
     save_path = Path(save_path)
     if save_path.suffix == '':
         # when suffix is already raw/bin/dat do not change it.
         save_path = save_path.parent / (save_path.name + '.dat')
 
-    if chunk_size is None:
+    if chunk_size is not None or chunk_mb is not None:
+        if time_axis == 1:
+            print("Chunking disabled due to 'time_axis' == 1")
```
samuelgarcia: Good idea.

samuelgarcia:
This is cool.
The way is written is a bit complicated.
the cascade of if else could only serve to get the chunk_size (int or None) and then there would be only one exolicit loop.
But it is ok.

alejoe91:
I agree :) I'll simplify it

samuelgarcia:
:)

PR #322, opened 2020-02-21
Bug when grp is not an int.

samuelgarcia:


samuelgarcia:
@alejoe91 : could you check this ?
This prevent plot_1_preprocessing example to work correctly.

alejoe91:
Hi Sam, I don't think that's correct because channel groups could be shuffled. Why is it failing with the other code? It's parsed as int

samuelgarcia:
I agree this is bad.
But look at **plot_1_preprocessing** in si please.

samuelgarcia:
Oups sorry this is **plot_3_sorting_by_group**

alejoe91:
Ahhhh! ok I think the problem is that groups can be `str` as well :)

alejoe91:
fixing here instead: https://github.com/SpikeInterface/spikesorters/pull/104

samuelgarcia:
We can close.

PR #324, opened 2020-02-23
Now saves also hs2 compatible unit/spike features if available in prop…

mhhennig:
…erties.


samuelgarcia:
OK for me.
@colehurwitz I let you check and merge.

colehurwitz:
Looks good

PR #329, opened 2020-02-24
keep_good_only False by default

alejoe91:
The `keep_good_only` param should be False by default. It's much easier to filter good units after than the opposite, since they are saved as `KSLabel` property. One can simply do:

```
good_units = []
for u in sort_ks.get_unit_ids():
    if sort_ks.get_unit_property(u, 'KSLabel') == 'good':
        good_units.append(u)

sort_good = se.SubSortingExtractor(sort_ks, unit_ids=good_units)
```

### DIFF
--- a/spikeextractors/extractors/kilosortextractors/kilosortextractors.py
+++ b/spikeextractors/extractors/kilosortextractors/kilosortextractors.py
@@ -28,7 +28,7 @@ class KiloSortSortingExtractor(PhySortingExtractor):
     is_writable = True
     mode = 'folder'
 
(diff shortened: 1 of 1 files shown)


alejoe91:
@samuelgarcia @mhhennig @colehurwitz do you guys agree? 

(on our data 'good' units are to few..we still prefer to manually curate with phy)

samuelgarcia:
We should ask authors there position about this because they will be benchmark with default option.

alejoe91:
This could be easily changed in the SF wrapper. But without returning all units it's much harder. I think that should be default for userd

samuelgarcia:
Do I merge this before the release ?

PR #331, opened 2020-03-06
Memmap

alejoe91:
This is to enable the creation of memmap arrays for sorting extractors for waveforms, pca scores, amplitudes, etc

- added self._tmp_folder (+set-get)
- modified set get features so that it does force conversion to in-memory numpy array

### DIFF
--- a/spikeextractors/sortingextractor.py
+++ b/spikeextractors/sortingextractor.py
@@ -1,6 +1,9 @@
 from abc import ABC, abstractmethod
 import numpy as np
 import copy
+import random
+import string
+import tempfile
 from pathlib import Path
 import shutil
 from .extraction_tools import get_sub_extractors_by_property
@@ -640,6 +643,8 @@ class SortingExtractor(ABC):
         temp_folder: Path
             The temporary folder
         '''
+        if self._tmp_folder is None:
+            self._tmp_folder = Path(tempfile.mkdtemp())
         return self._tmp_folder
 
     def set_tmp_folder(self, folder):
@@ -652,6 +657,36 @@ class SortingExtractor(ABC):
             The temporary folder
         '''
         self._tmp_folder = Path(folder)
+
+    def allocate_array(self, shape, dtype, memmap, name=None):
+        '''
+        Allocates a memory or memmap array
+
+        Parameters
+        ----------
+        shape: tuple
+            Shape of the array
+        dtype: dtype
+            Dtype of the array
+        memmap: bool
+            If True, array is memmaped
+        name: str
+            Name (root) of the file (if memmap is True)
+
+        Returns
+        -------
+        arr: np.array or np.memmap
+            The allocated memory or memmap array
+        '''
+        if memmap:
+            tmp_folder = self.get_tmp_folder()
+            if name is None:
+                name = ''.join([random.choice(string.ascii_letters) for i in range(5)])
+            arr = np.memmap(tmp_folder / (name + '.raw'), shape=shape, dtype=dtype)
+            arr[:] = 0
+        else:
+            arr = np.zeros(shape, dtype=dtype)
+        return arr
 
     @staticmethod
     def write_sorting(sorting, save_path):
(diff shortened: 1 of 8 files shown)


spikeextractors/sortingextractor.py
```diff
@@ -102,7 +112,7 @@ def set_unit_spike_features(self, unit_id, feature_name, value):
                 if unit_id not in self._unit_features.keys():
                     self._unit_features[unit_id] = {}
                 if isinstance(feature_name, str) and len(value) == len(self.get_unit_spike_train(unit_id)):
-                    self._unit_features[unit_id][feature_name] = np.asarray(value)
+                    self._unit_features[unit_id][feature_name] = value  # np.asarray(value) this is to keep memmap object
```
alejoe91: @colehurwitz any specific reason why this was cast as array? is it for lists?
I removed the cast to enable memmap arrays, but I can add a check for lists

colehurwitz: I don't think there was much reason for casting it. Since spiketrains are numpy arrays, I think I wanted to keep the 1-1 correspondence between features and spike trains. I can't remember any other reason why.

alejoe91: ok then I'll keep this way. As long as they have the same length, it's ok

colehurwitz:
So the memmaping won't happen in the SortingExtractor class, but it will accept memmaped objects basically? How does this work?

alejoe91:
> So the memmaping won't happen in the SortingExtractor class, but it will accept memmaped objects basically? How does this work?

This delas with dealing with the tmp_folder (deleted by the destructor). Temporary files can be saved in the `self._tmp_folder` by other functions. See this PR: https://github.com/SpikeInterface/spiketoolkit/pull/303

colehurwitz:
Hmm, do you think the logic for the memmap should be inside sortingextractor? So you can add features with "set_features" but specify memmap in that function?

alejoe91:
there is no logic here, it's just a container in which we can save memmaped stuff. And it's deleted when the object is destructed (that's why it's convenient to have it here)

colehurwitz:
Ok, your solution is good!

colehurwitz:
Ready to merge?

spikeextractors/sortingextractor.py
```diff
@@ -155,8 +165,13 @@ def get_unit_spike_features(self, unit_id, feature_name, start_frame=None, end_f
                             start_frame = 0
                         if end_frame is None:
                             end_frame = np.inf
-                        spike_indices =  np.where(np.logical_and(spike_train >= start_frame, spike_train < end_frame))
-                        return self._unit_features[unit_id][feature_name][spike_indices]
+                        if start_frame == 0 and end_frame == np.inf:
```
samuelgarcia: Having inf for int is a bad idea. None sounds better. (or -1 but more dangerous)

samuelgarcia:
I don't get the idea.
Where this self._tmp_folder will be used.
Why not having the same mechanism for recordingextractor and so the cacherecordingextractor would b ethe same ?

spikeextractors/sortingextractor.py
alejoe91: in this case it's fine because it's only used internally with the np.where (so -1 won't work)

samuelgarcia: Yes it is fine but we have to avoid this everywhere. No inf for indexing.

alejoe91:
Ok I think this is ready @samuelgarcia

alejoe91:
@samuelgarcia I added the creation of the folder and the allocate_array (also for the recordingextractor). I modified a bit the cache extractor to use the _tmp_folder.
Also in the allocate memory, one can initialize with an array (this will clean up spiketoolkit a lot!!)

samuelgarcia:
This is cool no ?

alejoe91:
yes I think we can merger this

PR #332, opened 2020-03-09
Inf epoch

colehurwitz:
@samuelgarcia @alejoe91 . So I am thinking that instead of infinity, end_frame can be set to None. If these changes are ok, I will propagate to spiketoolkit and release new versions soon.

### DIFF
--- a/spikeextractors/sortingextractor.py
+++ b/spikeextractors/sortingextractor.py
@@ -144,6 +144,7 @@ class SortingExtractor(ABC):
             An array containing all the features for each spike in the
             specified unit given the range of start and end frames.
         '''
+        start_frame, end_frame = self._cast_start_end_frame(start_frame, end_frame)
         if isinstance(unit_id, (int, np.integer)):
             if unit_id in self.get_unit_ids():
                 if unit_id not in self._unit_features.keys():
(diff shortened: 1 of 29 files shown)


spikeextractors/sortingextractor.py
```diff
@@ -486,7 +486,7 @@ def get_unit_spike_features(self, unit_id, feature_name, start_frame=None, end_f
                     self.set_unit_spike_features(unit_id=unit_id, feature_name=curr_feature_name, value=value)
 
     def add_epoch(self, epoch_name, start_frame, end_frame):
-        '''This function adds an epoch to your sorting extractor that tracks
+        '''This function adds an epoch to your recording extractor that tracks
```
alejoe91: this is the sorting extractor, no?

colehurwitz: Whoops. Good catch.

colehurwitz:
What do you guys think of this solution? A little annoying that I had to change all the extractors, but should be effective. No more infinities are allowed to be passed into extractors.

mhhennig:
Just checked this branch, as it is it does not solve the problem with `st.validation.compute_snrs` - still getting `OverflowError: cannot convert float infinity to integer` here:
```
...sortingextractor.py in _cast_start_end_frame(self, start_frame, end_frame)
    643             raise ValueError("start_frame must be an int, float (not infinity), or None")
    644         if isinstance(end_frame, (float, np.float)):
--> 645             end_frame = int(end_frame)
    646         elif isinstance(end_frame, (int, np.integer, type(None))):
    647             end_frame = end_frame
```

colehurwitz:
The spiketoolkit PR should fix that problem. That was a quality metric problem.

samuelgarcia:
Hi Cole,
OK for me.

Why not include the logic 
```python
        if start_frame is None:
            start_frame = 0
        if end_frame is None:
            ....
```
inside 
```self._cast_start_end_frame(start_frame, end_frame)```
and rename it to 
```self._ensure_start_end_frame(start_frame, end_frame)```

 ?

colehurwitz:
Some methods don’t convert end_frame to anything (subrecording, etc.) so I didn’t make it part of the function.

samuelgarcia:
OK I andesrtand.

colehurwitz:
I think this must be a spikeextractors memory map issue. @alejoe91

colehurwitz:
Should I merge?

alejoe91:
@mhhennig that is fixed with spikeextractors master. You can merge :)

PR #333, opened 2020-03-17
Mea1k implementation and MaxOne fixes

alejoe91:


samuelgarcia:
You could have done this directly in neo, no ?
THe more we implement here, the more we will have to move to neo...

alejoe91:
It's basically a copy of the MAxOne extractor which was already implemented. We still need a hackaton to port to neo anyways!

PR #334, opened 2020-03-17
Proposal for pickle load/dump

samuelgarcia:
Hi all.
@alejoe91 @colehurwitz @magland 

This is a proposal to make pickle/load for extactor.
Alessio have already merge a system that make possible to dump/load with json all extactors.
Including nested extactors for instance after a filter.

This is an extention to the previous PR to also use this system with pickle (usefull with joblib/multiprocessing for instance).

The main problem is that we need to hack the ```__init__``` in all extactor and replace with ```_init_extractor```.

Please have a look to baseextactor and BinDatRecordingExtractor that have no more init.

Then we can extend this to all extractors.

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -9,9 +9,17 @@ import datetime
 class BaseExtractor:
     is_dumpable = True
 
-    def __init__(self):
+    def __init__(self, *args, **kargs):
         self._kwargs = {}
         self._tmp_folder = None
+        
+        self.
+    
+    def _init_extractor(self, *args, **kargs):
+        # this remplace the init in each class
+        # this is call both at __init__ but also in __setstate__ (when pickle.load)
+        raise NotImplementedError
+    
 
     def make_serialized_dict(self):
         class_name = str(type(self)).replace("<class '", "").replace("'>", '')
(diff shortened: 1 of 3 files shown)


spikeextractors/baseextractor.py
```diff
@@ -9,9 +9,17 @@
 class BaseExtractor:
     is_dumpable = True
 
-    def __init__(self):
+    def __init__(self, *args, **kargs):
         self._kwargs = {}
         self._tmp_folder = None
+        
+        self.
```
alejoe91: ???

tests/test_extractors.py
```diff
@@ -424,13 +426,20 @@ def _check_sortings_equal(self, SX1, SX2):
...
         extractor.dump(file_name='test.json')
         extractor_loaded = se.load_extractor_from_json('test.json')
 
         if 'Recording' in str(type(extractor)):
             self._check_recordings_equal(extractor, extractor_loaded)
         elif 'Sorting' in str(type(extractor)):
             self._check_sortings_equal(extractor, extractor_loaded)
+        
+        
+        # with pickle
```
alejoe91: this will fail for all non implemented extractors. Maybe you can make another function and we add it gradually as we add the __init__extractor functions?

alejoe91:
@samuelgarcia can we close this? I think that the current version is good :)

PR #338, opened 2020-03-27
Proposal: get_traces decorator

alejoe91:
@colehurwitz already added some checks for parsing start and end frame.
I suggest a decorator in the extractor tools that:
- checks `start_frame` and `end_frame`
- checks that all `channel_ids` are correct / filters out wrong ones
- parses to `start_frame` and `end_frame` 

Note that this can be accessed by `spiketoolkit` and could be extended in the future. It would cleanup a lot of the code and fix unwanted bugs (e.g. passing `start_frame` < 0 was not caught before o it needed to be done in each get_traces). 

The `@wraps` decorator makes sure that the docstrings is kept.
If you give me the green light I'll do some cleanup :)

### DIFF
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -604,22 +604,23 @@ def check_get_traces_args(func):
             channel_ids = recording.get_channel_ids()
         if start_frame is not None:
             if start_frame < 0:
-                print("'start_time' set to", 0)
-                start_frame = 0
+                start_frame = recording.get_num_frames() - start_frame
         else:
             start_frame = 0
         if end_frame is not None:
             if end_frame > recording.get_num_frames():
                 print("'end_time' set to", recording.get_num_frames())
                 end_frame = recording.get_num_frames()
+            elif end_frame < 0:
+                end_frame = recording.get_num_frames() - end_frame
         else:
             end_frame = recording.get_num_frames()
-        assert end_frame - start_frame > 0, "'start_frame' is equal of greater than 'end_frame'!"
-        _cast_start_end_frame(recording, start_frame, end_frame)
+        assert end_frame - start_frame > 0, "'start_frame' must be less than 'end_frame'!"
+        cast_start_end_frame(start_frame, end_frame)
         kwargs['channel_ids'] = channel_ids
         kwargs['start_frame'] = start_frame
         kwargs['end_frame'] = end_frame
-        
+
         # pass recording as arg and rest as kwargs
         get_traces_correct_arg = func(args[0], **kwargs)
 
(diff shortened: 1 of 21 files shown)


spikeextractors/extraction_tools.py
```diff
@@ -562,4 +562,82 @@ def load_extractor_from_dict(d):
...
\ No newline at end of file
+    return BaseExtractor.load_extractor_from_dict(d)
+
+
+def check_get_traces_args(func):
+    @wraps(func)
+    def corrected_args(*args, **kwargs):
+        # parse args and kwargs
+        if len(args) == 1:
+            recording = args[0]
```
samuelgarcia: why do you play with arg wheras get_traces is kargs ?

samuelgarcia:
THis is OK for me.
Just one comments.

bendichter:
What checks would it perform? Checking that `start_frame` and `end_frame` are `int`s makes sense, but the downstream error of indexing with non-ints is pretty striaghtforward. 

I don't know about checking that they are positive. I might expect users to intentionally use a negative `start_frame` or `end_frame`, and expects standard `numpy` indexing, i.e. "this many indices from the last index"

alejoe91:
> THis is OK for me.
> Just one comments.

If you call `rec.get_traces([0, 1], 0, 100)`, `channel_ids`, `start_time` and `end_time` are `args`

alejoe91:
> What checks would it perform? Checking that `start_frame` and `end_frame` are `int`s makes sense, but the downstream error of indexing with non-ints is pretty striaghtforward.
> 
> I don't know about checking that they are positive. I might expect users to intentionally use a negative `start_frame` or `end_frame`, and expects standard `numpy` indexing, i.e. "this many indices from the last index"

Hi Ben,
 I see your point. I don't think that all extractors support `np.array` indexing. What I can do is checking that `start_frame` < `end_frame` (that should alwas be true) and if they are negative set them to e.g. `rec.get_num_frames() - end_frame`. Sounds good?

bendichter:
@alejoe91 sounds good to me

alejoe91:
> Hi Ben,
> I see your point. I don't think that all extractors support `np.array` indexing. What I can do is checking that `start_frame` < `end_frame` (that should alwas be true) and if they are negative set them to e.g. `rec.get_num_frames() - end_frame`. Sounds good?

I meant `rec.get_num_frames() + end_frame`

spikeextractors/extraction_tools.py
alejoe91: it can be args as well. there is no `*`

alejoe91: It could be args as well as there is no `*`

alejoe91:
Allright fellas merging this

PR #340, opened 2020-03-29
is_dumpable is an object variable, not a statc variable

alejoe91:
This is to accomodate the `NumpyRecordingExtractor`, which can be dumpable (if `timeseries` is a npy file), or not dumpable (if `timeseries` is a numpy array)

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -8,11 +8,10 @@ from copy import deepcopy
 
 
 class BaseExtractor:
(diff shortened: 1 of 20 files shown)


samuelgarcia:
OK.
It could even be a method.

alejoe91:
it's already a method. It checks that everything downstream is dumpable `BaseExtractor.check_if_dumpable()`

samuelgarcia:
cool.

PR #342, opened 2020-03-30
Add `is_filtered` to BaseExtractor

alejoe91:
This makes it easy to know if a RecordingExtractor is filtered or not (especially for spiketoolkit)

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -12,6 +12,7 @@ class BaseExtractor:
         self._kwargs = {}
         self._tmp_folder = None
         self.is_dumpable = True
+        self.is_filtered = False
 
     def make_serialized_dict(self):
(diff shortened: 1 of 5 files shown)


alejoe91:
@colehurwitz this is useful for not re-filtering in validation and curation if recordings are already filtered

samuelgarcia:
Good idea. Maybe it could be more formal and have list of "done preprocessing" (like whittening) that would also contain a dict with args of theses steps.

samuelgarcia:
Or changed this attributes self.is_filtered to self.is_preprocessed
This is more generic and less ambiguous.

alejoe91:
I think that `is_filtered` is of particular importance and stronger than `is_processed`. Currently is_filtered is only changed by the `bandpass_filter`. We can think about other more generic approaches if needed in the future.

PR #346, opened 2020-03-31
removed gui params and exporter terminology

colehurwitz:
As discussed with @alejoe91 , I plan to move all GUI logic to spikely to disentangle GUI logic from spikeinterface.


alejoe91:
:heart:

PR #349, opened 2020-04-01
Make neo soft dependency real

alejoe91:


samuelgarcia:
Why ?
Don't we agree to move slowly towards neo ?

alejoe91:
you said yesterday that neo should be a soft dependency! I agree when we make the move it'll be a strong dependency. I can leave it strong if you prefer :) maybe I misunderstood

PR #351, opened 2020-04-03
Fixes to bindat, maxone, mea1k extractors

alejoe91:
- found a bad bug in the `read_binary` function, was affecting performance A LOT. Basically, the memmap file was open and closed within the function...

- fixed some indexing in MAxOne and Mea1k extractors

- Cache creates folder if non-existing


colehurwitz:
Wow, good catch

alejoe91:
in extracting waveforms, it increases performance 100x...it was easy to spot :P there was also a rpblem with the indexing! I'll merge this

PR #354, opened 2020-04-08
klusta extractor; add 'cluster_group' and 'cluster_group_name' as unit properties

TRuikes:
Here my implementation of adding 'cluster_group' and 'cluster_group_name' as unit properties.

I slightly updated the way of reading the data from the .kwik files, as I think this way it's easier to see wherefrom data in the hdf5 tree is loaded.

### DIFF
--- a/spikeextractors/extractors/klustaextractors/klustaextractors.py
+++ b/spikeextractors/extractors/klustaextractors/klustaextractors.py
@@ -111,7 +111,7 @@ class KlustaSortingExtractor(SortingExtractor):
             for cluster_id in chan_cluster_ids:
                 cluster_frame_idx = np.nonzero(chan_cluster_id_arr == cluster_id)  # the [()] is a h5py thing
                 st = chan_cluster_times_arr[cluster_frame_idx]
-                assert st.shape[0] > 0, 'no spikes in cluster'  # this shouldnt happen
+                assert st.shape[0] > 0, 'no spikes in cluster'
                 cluster_group = kf_reader.get(f'/channel_groups/{channel_group}/clusters/main/{cluster_id}').attrs['cluster_group']
 
                 assert cluster_group in self.default_cluster_groups.keys(), f'cluster_group not in "default_dict: {cluster_group}'
--- a/spikeextractors/extractors/klustaextractors/klustaextractors.py
+++ b/spikeextractors/extractors/klustaextractors/klustaextractors.py
@@ -59,7 +59,7 @@ class KlustaSortingExtractor(SortingExtractor):
 
     default_cluster_groups = {0: 'Noise', 1: 'MUA', 2: 'Good', 3: 'Unsorted'}
 
-    def __init__(self, file_or_folder_path):
+    def __init__(self, file_or_folder_path, exclude_cluster_groups=None):
         assert HAVE_KLSX, "To use the KlustaSortingExtractor install h5py: \n\n pip install h5py\n\n"
         SortingExtractor.__init__(self)
         kwik_file_or_folder = Path(file_or_folder_path)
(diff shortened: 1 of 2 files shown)


spikeextractors/extractors/klustaextractors/klustaextractors.py
```diff
@@ -37,13 +49,16 @@ def __init__(self, folder_path):
...
 
+# noinspection SpellCheckingInspection
 class KlustaSortingExtractor(SortingExtractor):
     extractor_name = 'KlustaSortingExtractor'
     installed = HAVE_KLSX  # check at class level if installed or not
     installation_mesg = "To use the KlustaSortingExtractor install h5py: \n\n pip install h5py\n\n"  # error message when not installed
     is_writable = True
     mode = 'file_or_folder'
 
+    default_cluster_groups = {0: 'Noise', 1: 'MUA', 2: 'Good', 3: 'Unsorted'}
```
alejoe91, line 60: this is fixed in the kwik file, right? 

Then I recommend only saving the `cluster_groups_name` and not the `.cluster_groups_int`. Do you agree?

spikeextractors/extractors/klustaextractors/klustaextractors.py
```diff
@@ -69,41 +84,48 @@ def __init__(self, file_or_folder_path):
...
+
         if len(np.unique(klusta_units)) == len(np.unique(unique_units)):
             self._unit_ids = klusta_units
         else:
             print('Klusta units are not unique! Using unique unit ids')
             self._unit_ids = unique_units
         for i, u in enumerate(self._unit_ids):
             self.set_unit_property(u, 'group', groups[i])
+            self.set_unit_property(u, 'cluster_group', cluster_groups_int[i])
+            self.set_unit_property(u, 'cluster_group_name', cluster_groups_name[i])
```
alejoe91, line 125: Would you mind calling the property `quality` so it's the same as the `phy` extractor?

spikeextractors/extractors/klustaextractors/klustaextractors.py
```diff
@@ -37,13 +49,16 @@ def __init__(self, folder_path):
...
+# noinspection SpellCheckingInspection
 class KlustaSortingExtractor(SortingExtractor):
     extractor_name = 'KlustaSortingExtractor'
     installed = HAVE_KLSX  # check at class level if installed or not
     installation_mesg = "To use the KlustaSortingExtractor install h5py: \n\n pip install h5py\n\n"  # error message when not installed
     is_writable = True
     mode = 'file_or_folder'
 
+    default_cluster_groups = {0: 'Noise', 1: 'MUA', 2: 'Good', 3: 'Unsorted'}
+
```
alejoe91, line 61: It could be an option to have an `exclude_cluster_groups` argument similar to the [PhySortingExtractor](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/phyextractors/phyextractors.py#L59). This can be used to exclude noise or unsorted units automatically (see [here](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/phyextractors/phyextractors.py#L131-L140)). What do you think?

alejoe91:
Thanks @TRRuikes and good job on your first PR :) I have a few comments, we can discuss about those

spikeextractors/extractors/klustaextractors/klustaextractors.py
TRuikes: No these are the gui's defaults. I think you can also add manual labels, but I/our lab doesn't use this feature, so I don't know how it would look like. I would expect a new group integer (ie '4'), which stands for the custom label.

With that context, how would you approach this?

spikeextractors/extractors/klustaextractors/klustaextractors.py
TRuikes: will do

spikeextractors/extractors/klustaextractors/klustaextractors.py
TRuikes: yes that's useful, added it

spikeextractors/extractors/klustaextractors/klustaextractors.py
TRuikes: I've added this:
`                assert cluster_group in self.default_cluster_groups.keys(), f'cluster_group not in "default_dict: {cluster_group}'`

if that will ever fail for someone, we can pick it up from there

alejoe91: That could work

alejoe91:
Hi @TRuikes 

Tests are failing because the `write_sorting` was not updated. Two options: either you modify the write sorting and write the required info, or you delete the `write_sorting` and remove the test. I think the second option is fine, as I don't think it will be of great use

TRuikes:
@alejoe91 when you say 'tests are failing', do you mean Travis on this page? The push without the write method fails tough...

In `master`, when I use the write method, it writes invalid files, as `sample_rate` is not written as an attribute. I'm not too familiar with the exact kwik model. I think it's easier to remove the method completely (the format will be depracted i think anyway?).

alejoe91:
Hi @TRuikes 

They still fail because you should remove [this test](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/tests/test_extractors.py#L236-L242) as well. Is you remove that tests should pass. 
I agree it's ok to remove it

TRuikes:
Allright! thanks for the input

PR #356, opened 2020-04-14
units spike train implemented

colehurwitz:
@alejoe91

### DIFF
--- a/spikeextractors/sortingextractor.py
+++ b/spikeextractors/sortingextractor.py
@@ -75,6 +75,28 @@ class SortingExtractor(ABC, BaseExtractor):
         '''
         pass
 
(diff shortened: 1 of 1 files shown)


spikeextractors/sortingextractor.py
```diff
@@ -75,6 +75,28 @@ def get_unit_spike_train(self, unit_id, start_frame=None, end_frame=None):
         '''
         pass
 
+    def get_units_spike_train(self, unit_ids=None, start_frame=None, end_frame=None):
```
samuelgarcia, line 78: I am wondering if the naming is self explaining.
And maybe having a method that differ only woth  a plurial can be missleading no ?

colehurwitz: That is usually our naming convention. `get_units_property` is another example. Do you think we should make it more clear?

PR #357, opened 2020-04-15
Refractor properties and features for dumping

alejoe91:
Final dumps! I think this is super important to save the status of recording and sortings (e.g. after extracting waveforms, pca, etc...)

- Added the consept of `key_properties` that for RecordingExtractors are `group` and `location`

- This cleans up property and feature handling, now done at the base class (`self._properties` and `self._features`).

- `make_serialized_dict()` allows saves `key_properties` by default

- 3 dump functions: `dump_to_dict`, `dump_to_json` (both save key properties) and `dump_to_pickle`, that enable saving all features and properties

- The `load_extractor_from_dict` and `load_extractor_from_json` have been modified to load the properties and features and the `load_extractor_from_pickle` has been added

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -55,7 +55,7 @@ class BaseExtractor:
         '''
         return self.make_serialized_dict()
 
-    def dump_to_json(self, file_path=None, dump_properties_and_features=True):
+    def dump_to_json(self, file_path=None):
         '''
         Dumps recording extractor to json file.
         The extractor can be re-loaded with spikeextractors.load_extractor_from_json(json_file)
@@ -64,8 +64,6 @@ class BaseExtractor:
         ----------
         file_path: str
             Path of the json file
-        dump_properties_and_features: bool
-            If True, property and features are dumped as npz files in the same folder as the json
         '''
         if self.check_if_dumpable():
             if file_path is None:
@@ -95,8 +93,10 @@ class BaseExtractor:
         ----------
         file_path: str
             Path of the json file
-        dump_properties_and_features: bool
-            If True, property and features are dumped as npz files in the same folder as the json
+        include_properties: bool
+            If True, all properties are dumped
+        include_features: bool
+            If True, all features are dumped
         '''
         if self.check_if_dumpable():
             dump_dict = {}
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -6,16 +6,19 @@ import numpy as np
 import datetime
 from copy import deepcopy
 import tempfile
+import pickle
 
 
 class BaseExtractor:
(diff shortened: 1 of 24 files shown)


spikeextractors/baseextractor.py
```diff
@@ -32,47 +39,47 @@ def make_serialized_dict(self):
...
-                    dump_dict['groups'] = groups
         else:
             dump_dict = {'class': class_name, 'module': module, 'kwargs': {},
                          'version': imported_module.__version__, 'dumpable': False}
-
+        dump_dict = self._include_properties_and_features_in_dict(dump_dict, include_properties, include_features)
         return dump_dict
 
-    def dump(self, file_name=None, folder_path=None):
+    def dump(self, file_path=None, dump_properties_and_features=True):
```
samuelgarcia, line 48: I think I would put the same signautre than ```make_serialized_dict``` with include_properties=None, include_features=None.
But by default I would put include_properties='all' and 'all would be all feature or properties.

samuelgarcia:
I don't known what to think.
This dump is not as elegant as the previous one with only json.

I would propose a dump_json (as before) and a dump that would use only pickle include for the dict.
This hybrid json/pickle don't make me so happy.
Even I understand the need.

alejoe91:
I know, but we need to dump data as well...I don't want to compute waveforms more than one time and we also make analysis that we can save as property/fetaure and reload later. Any other suggestion?

samuelgarcia:
My suggestion would be:
  * keep of course the serilized to dict.
  * dump use pickle and only pickle for dict and properties but in the same file.
  * a separate dump_json with no properties.

It would be more coherent no ?

alejoe91:
> My suggestion would be:
> 
> * keep of course the serilized to dict.
> * dump use pickle and only pickle for dict and properties but in the same file.
> * a separate dump_json with no properties.
> 
> It would be more coherent no ?

So then one would need to call 2 functions to save and 2 functions to load the data? Or you suggest to dump 1) dict, 2) properties, 2) features in the same pickle?

alejoe91:
so then the dump would save a pickle and the dump_json would save a json..that can work

samuelgarcia:
I think that **_key_properties** is not the appropriate naming.

alejoe91:
Apart from that, I think this is cool :)

spiketoolkit is ready as well

alejoe91:
@samuelgarcia this is ready to be merged

samuelgarcia:
I think the week end break my chance to review this.
Have you propagate to spikecomparison.load_probe_file_inplace ?
This is quite important for GTstudy.

alejoe91:
Ahhaah hi Sam! The API is the same and all tests from other packages are fine :) you can review it now and I can make extra changes, but this is exactly what we agreed on! (Except the naming)

PR #363, opened 2020-04-20
BF: do not assume in __del__ that ._tmp_folder is available

yarikoptic:
For a variety of reasons, one of which is that it is typical with python3 to arrive
to `__del__` whenever other attributes were freed already.  It is a bigger issue I am
not going to solve here, rather just to prevent a non interformative crash

### DIFF
--- a/spikeextractors/recordingextractor.py
+++ b/spikeextractors/recordingextractor.py
@@ -22,7 +22,7 @@ class RecordingExtractor(ABC, BaseExtractor):
         self.id = random.randint(a=0, b=9223372036854775807)
 
     def __del__(self):
-        if self._tmp_folder is not None:
+        if getattr(self, '_tmp_folder', None):
             try:
                 shutil.rmtree(self._tmp_folder)
             except Exception as e:


alejoe91:
Hi @yarikoptic 

Does the get_attr function check if _tmp_folder is not None? Sorry I never used it before :)

yarikoptic:
> Hi @yarikoptic
> 
> Does the get_attr function check if _tmp_folder is not None? Sorry I never used it before :)

it doesn't ... `None` (in `if`) evaluates to False. So either it is assigned and None or an empty string, or it is not assigned and thus the default value `None` is used, .... HA! I (you) found a bug in my code, it should have been `if not getattr(self, '_tmp_folder', None):` ;) I will force push the corrected code in a second. thanks for the question ;)

yarikoptic:
nah... d'oh -- I should put this laptop aside for today.  code was correct.  So, if it is not assigned anyhow -- it would be None and nothing would be removed. I am pushing back correct code ;)

PR #370, opened 2020-04-21
Fix str in dumps

alejoe91:
Fixes dumps when file names are not passed

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -67,9 +67,9 @@ class BaseExtractor:
         '''
         if self.check_if_dumpable():
             if file_path is None:
-                if 'Recording' in str(self.__class__):
+                if 'Recording' in type(self).__name__:
                     file_path = 'spikeinterface_recording.json'
-                elif 'Sorting' in str(self.__class__):
+                elif 'Sorting' in type(self).__name__:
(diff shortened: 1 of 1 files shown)


spikeextractors/baseextractor.py
```diff
@@ -67,9 +67,9 @@ def dump_to_json(self, file_path=None):
         '''
         if self.check_if_dumpable():
             if file_path is None:
-                if 'Recording' in self.__class__:
+                if 'Recording' in str(self.__class__):
```
samuelgarcia, line 70: or self.__class__.__name__

alejoe91: that doesn't work

alejoe91: `type(self).__name__` works as well, or `str(self.__class__)`

alejoe91: same as thos type(self).__name__

yarikoptic:
Thanks! I will check when get to the laptop. I hope you don't mind me commenting: it feels like there is just missing class method (or even just an attribute) defined at the top base classes like `get_default_filename` to avoid all the if, and just to rely on inheritance.

alejoe91:
fixed here https://github.com/SpikeInterface/spikeextractors/pull/372

PR #371, opened 2020-04-21
A sorting extractor for JRCLUST output

aliddell:
Reads pertinent data from the JRCLUST output mat file. On the way I implemented a rudimentary .mat file reader that should handle both old-style and new-style .mat files. Writing output is tricky because I'm not sure how one could write a .mat file, as opposed to an HDF5 file.

Let me know if any naming conventions are off!

### DIFF
--- a/spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
+++ b/spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
@@ -1,0 +1,96 @@
+from pathlib import Path
+from typing import Union
+
+import numpy as np
+
+from spikeextractors.extractors.matsortingextractor.matsortingextractor import MATSortingExtractor, HAVE_MAT
+from spikeextractors.extraction_tools import check_valid_unit_id
+
+PathType = Union[str, Path]
+
+
+class JRCSortingExtractor(MATSortingExtractor):
+    extractor_name = "JRCSortingExtractor"
+    installation_mesg = "To use the MATSortingExtractor install h5py and scipy: \n\n pip install h5py scipy\n\n"  # error message when not installed
+
+    def __init__(self, file_path: PathType, positive_only: bool = False):
+        super().__init__(file_path)
+        file_path = self._kwargs["file_path"]
+
+        spike_times = self._getfield("spikeTimes").ravel() - 1  # int32
+        spike_clusters = self._getfield("spikeClusters").ravel()  # uint32
+        spike_amplitudes = self._getfield("spikeAmps").ravel()  # int16
+        spike_sites = self._getfield("spikeSites").ravel()  # uint32
+        spike_positions = self._getfield("spikePositions").T  # float32
+
+        unit_centroids = self._getfield("clusterCentroids").astype(np.float).T
+        unit_sites = self._getfield("clusterSites").astype(np.uint32).ravel()
+        mean_waveforms = self._getfield("meanWfGlobal").T
+        mean_waveforms_raw = self._getfield("meanWfGlobalRaw").T
+
+        # try to extract the sample rate from the .prm file
+        sample_rate = None
+        prm_file = Path(file_path.name.replace("_res.mat", ".prm"))
+        with prm_file.open("r") as fh:
+            line = fh.readline()
+            while line:
+                line = line.split('%')[0].strip(" ;")
+                if line.startswith("sampleRate"):
+                    line = line.split("=", 1)
+                    try:
+                        sample_rate = float(line[1])
+                    except (IndexError, ValueError):
+                        raise AttributeError("Malformed or missing value for sample rate.")
+                    break
+                line = fh.readline()
+
+        if sample_rate is None:
+            raise AttributeError("Sample rate was not defined.")
+
+        self.set_sampling_frequency(sample_rate)
+
+        # traces, features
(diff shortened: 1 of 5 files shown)


alejoe91:
Hi @aliddell 

Thanks! It looks great! Will give a closer look tomorrow and merge :)

spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
```diff
@@ -0,0 +1,96 @@
...
+        for uid in self._unit_ids:
+            mask = (spike_clusters == uid)
+            self._spike_trains[uid] = spike_times[mask]
+
+            self.set_unit_spike_features(uid, "spike_amplitudes", spike_amplitudes[mask])
+            self.set_unit_spike_features(uid, "spike_sites", spike_sites[mask])
+            self.set_unit_spike_features(uid, "spike_positions", spike_positions[mask, :])
+
+            self.set_unit_property(uid, "centroid", unit_centroids[uid - 1, :])
+            self.set_unit_property(uid, "center_site", unit_sites[uid - 1])
```
alejoe91, line 81: in the `spiketoolkit.postprocessing` module we call the same properties in this way:
`spike_amplitudes` -> `amplitudes`
`center_site` -> `max_channel`
`mean_waveform` - > `template`
( So I would use `template_raw` as well)

I would also drop the `spike_` before the other features, as they are already spike features

spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
```diff
@@ -0,0 +1,96 @@
...
+            self.set_unit_property(uid, "mean_waveform", mean_waveforms[:, :, uid - 1])
+            self.set_unit_property(uid, "mean_waveform_raw", mean_waveforms_raw[:, :, uid - 1])
```
alejoe91, line 83: Also, can you add this line?
`self._kwargs = {'file_path': str(Path(file_path).absolute(), 'keep_good_only': keep_good_only}`

This makes sure that the extractor can be correctly dumped to file and reloaded

spikeextractors/extractors/matsortingextractor/matsortingextractor.py
```diff
@@ -0,0 +1,56 @@
...
+
+from spikeextractors import SortingExtractor
+
+PathType = Union[str, Path]
+
+
+class MATSortingExtractor(SortingExtractor):
+    extractor_name = "MATSortingExtractor"
+    installed = HAVE_MAT  # check at class level if installed or not
+    is_writable = True
```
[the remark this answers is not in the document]
aliddell: Ah, I was wondering what that parameter was for.

spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
aliddell: Question: `center_site` in my implementation applies to the unit as a whole. What about for the individual spikes?

spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
aliddell: `file_path` is already in `MatSortingExtractor._kwargs`, but I'll add `keep_good_only`.

spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
alejoe91: What I'm saying is that already by setting a `spike_feature` instead of a `property` you are telling the user these are spike_sites, and not global site. So I would just drop the `spike_` part :)

aliddell: So you recommend calling them both `center_site`?

aliddell: Er, sorry, `max_channel`?

alejoe91: Yes :)

alejoe91: Of course if you agree!

spikeextractors/extractors/jrcsortingextractor/jrcsortingextractor.py
```diff
@@ -73,14 +73,16 @@ def __init__(self, file_path: PathType, positive_only: bool = False):
             mask = (spike_clusters == uid)
             self._spike_trains[uid] = spike_times[mask]
 
-            self.set_unit_spike_features(uid, "spike_amplitudes", spike_amplitudes[mask])
-            self.set_unit_spike_features(uid, "spike_sites", spike_sites[mask])
-            self.set_unit_spike_features(uid, "spike_positions", spike_positions[mask, :])
+            self.set_unit_spike_features(uid, "amplitudes", spike_amplitudes[mask])
+            self.set_unit_spike_features(uid, "max_channel", spike_sites[mask])
```
alejoe91, line 77: Sorry to bother again, maybe this can be max_channels for even more consistency? :)

alejoe91:
Thanks again! Merging :)

PR #372, opened 2020-04-21
RF+BF: centralize logic for handling dumping extractors into files

yarikoptic:
- avoid if/elif matching based on the class name (which might
  not even have those terms) and would rely on class inheritance to decide
  on default filename.

- avoid code duplication by concentrating it in _get_file_path which would
  do all necessary steps uniformly regardless or which them method invokes it.

- raises a dedicated NotDumpableExtractorError now instead of too generic
  Exception if for some reason current instance is not dumpable.

An alternative to #370 .  Completely alternatively, may be dumping to some default filename should be completely refactored out - IMHO better be explicit than implicit ;) (Zen of Python IIRC)

TODOs:
- [ ] a dedicated unittest which verifies storing into a default file path and that this new dedicated exception is raised

I have not approached the test since forgot good basic `unittest` with too much of the nose and pytest.  I see that pytest is used as a running on travis but not really used to compose the tests.  That is why, I would appreciate if someone takes over to finalize this PR (if you like it)

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -8,8 +8,16 @@ from copy import deepcopy
 import tempfile
 import pickle
 
+from .exceptions import NotDumpableExtractorError
+
 
 class BaseExtractor:
+
+    # To be specified in concrete sub-classes
+    # The default filename (extension to be added by corresponding method)
+    # to be used if no file path is provided
+    _default_filename = None
+
     def __init__(self):
         self._kwargs = {}
         self._tmp_folder = None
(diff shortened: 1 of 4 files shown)


alejoe91:
Thanks for this! I think it looks good. I can merge and implement the tests later afterwards

yarikoptic:
Up to you! I would have implemented a test first (you should be about to push to my branch and this update this PR) - what if I broke something? ;-)

alejoe91:
I don't think I can push to this PR, but I'll try

Issue #369, opened 2020-04-21
Neuralynx: ss.run_klusta() crashes on Neuralynx data due to broadcasting issue


yarikoptic:
NB sorry for the flood

```python
ss.run_klusta(recording)
```
leads to
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-cb71611d47a1> in <module>
----> 1 ss.run_klusta(recording)

~/anaconda-5.2.0-2.7/envs/mvdm-nept/lib/python3.7/site-packages/spikesorters/sorterlist.py in run_klusta(*args, **kwargs)
    195         The spike sorted data
    196     """
--> 197     return run_sorter('klusta', *args, **kwargs)
    198 
    199 

~/anaconda-5.2.0-2.7/envs/mvdm-nept/lib/python3.7/site-packages/spikesorters/sorterlist.py in run_sorter(sorter_name_or_class, recording, output_folder, delete_output_folder, grouping_property, parallel, verbose, raise_error, **params)
     79                          parallel=parallel, verbose=verbose, delete_output_folder=delete_output_folder)
     80     sorter.set_params(**params)
---> 81     sorter.run(raise_error=raise_error)
     82     sortingextractor = sorter.get_result()
     83 

~/anaconda-5.2.0-2.7/envs/mvdm-nept/lib/python3.7/site-packages/spikesorters/basesorter.py in run(self, raise_error)
    119     def run(self, raise_error=True):
    120         for i, recording in enumerate(self.recording_list):
--> 121             self._setup_recording(recording, self.output_folders[i])
    122 
    123         # dump again params because some sorter do a folder reset (tdc)

~/anaconda-5.2.0-2.7/envs/mvdm-nept/lib/python3.7/site-packages/spikesorters/klusta/klusta.py in _setup_recording(self, recording, output_folder)
     78             raw_filename = output_folder / 'recording.dat'
     79             dtype = 'int16'
---> 80             recording.write_to_binary_dat_format(raw_filename, time_axis=0, dtype=dtype, chunk_mb=500)
     81 
     82         if p['detect_sign'] < 0:

~/proj/misc/spikeextractors/spikeextractors/recordingextractor.py in write_to_binary_dat_format(self, save_path, time_axis, dtype, chunk_size, chunk_mb)
    715         '''
    716         write_to_binary_dat_format(self, save_path=save_path, time_axis=time_axis, dtype=dtype, chunk_size=chunk_size,
[...]

PR #375, opened 2020-04-22
Fixes to neo additional_gain

alejoe91:
fixes this https://github.com/SpikeInterface/spikeextractors/issues/369 the right way

### DIFF
--- a/spikeextractors/extractors/neoextractors/neobaseextractor.py
+++ b/spikeextractors/extractors/neoextractors/neobaseextractor.py
@@ -85,7 +85,7 @@ class NeoBaseRecordingExtractor(RecordingExtractor, _NeoBaseExtractor):
                                                                     channel_ids=channel_ids)
         channel_idxs = np.array([list(channel_ids).index(ch) for ch in channel_ids])
         # and then to uV
-        scaled_traces *= self.additional_gain[0, channel_idxs]
+        scaled_traces *= self.additional_gain[:, channel_idxs]
 
         # fortunatly neo works with (samples, channels) strides
         # so transpose to spieextractors wolrd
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/neoextractors/neobaseextractor.py
```diff
@@ -83,8 +83,9 @@ def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
         scaled_traces = self.neo_reader.rescale_signal_raw_to_float(raw_traces, dtype='float32',
                                                                     channel_indexes=None, channel_names=None,
                                                                     channel_ids=channel_ids)
+        channel_idxs = np.array([list(channel_ids).index(ch) for ch in channel_ids])
         # and then to uV
-        scaled_traces *= self.additional_gain
+        scaled_traces *= self.additional_gain[0, channel_idxs]
```
samuelgarcia, line 88: The ```[0, ``` is wrong this first dimention is needed to force the broacasting when there is ambiguity on dims.
So ```[:, ```

And to speed up this I would put 
```
If channel_ids is None:
    additional_gain = self.additional_gain
else:
    channel_idxs = np.array([list(channel_ids).index(ch) for ch in channel_ids])
    additional_gain = self.additional_gain[:, channel_idxs]
scaled_traces *= additional_gain[:, channel_idxs]
```

alejoe91: Channel IDs is never none due to the decoratorm you can push the changes you'd like, but additional_gain has a shape (1, num_channels)

samuelgarcia: I think it is a bad idea that channel_ids is never None.

Because when you need slice(None) to get a full dimenssion is really really faster that doing
array.take(...).

I don't known when this was done but forcing np.arange(nb_channel) instead of slice(None) is really really a perf killer to fetch data on a memap for instance.

alejoe91: no extractors slice except Biocam, and in that case the slicing is done in place (if channel ids are ordered and with no gaps)

PR #383, opened 2020-05-04
Mearec slice

alejoe91:


samuelgarcia:
Hi Alessio.
I think it would be faster to load the chunk sliced in time in memory and then slice in channel.

Like this:
```
time_chunk = self._recordings[:, start_frame:end_frame]
out = time_chunk[np.sort(channel_ids), :]
```


could you try it ?

alejoe91:
Yes I'll try later!

alejoe91:
@samuelgarcia your solution is slower. Slicing only is about 30% faster

alejoe91:
Actually, @mhhennig suggested to use lists instead of pre-allocating np.arrays in `get_snippets` and this makes it a 10% faster!

I'll add it in this PR

samuelgarcia:
Sorry. I thought it could the case.
On memmap, I think it is better.
Did you test this on ssd ?
hdf5 is a special case because it is block coded.

About list+concatenate: I am not sure of this because needs tiwe memory.
Normally preallocate is faster. 

The other big change as I said before is to not loop by unit_id because we run througth the entire file several times.

PR #386, opened 2020-05-06
remove multiple return type

colehurwitz:
@alejoe91 I really think we should not be returning different types (int or list) for any function. I just caught a bug in SpikeInterface where it was expecting a list and got an int. We always will need `if` statements everywhere if we do this to deal with ints or lists. We removed this in epochs as well.

### DIFF
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -203,7 +203,7 @@ def save_to_probe_file(recording, probe_file, grouping_property=None, radius=Non
         with probe_file.open('w') as f:
             if 'location' in recording.get_shared_channel_property_names():
                 for chan in recording.get_channel_ids():
-                    loc = recording.get_channel_locations(chan)
+                    loc = recording.get_channel_locations(chan)[0]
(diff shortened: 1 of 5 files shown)


colehurwitz:
I need to fix the tests still, but I think this will be helpful.

alejoe91:
Hi @colehurwitz 

Where did you find the bug? I think it's nice that if you pass an int you get an int (or a 1d array for locations), if you pass a list you get a list.

colehurwitz:
I found the bug in the spikeinterface examples.

What you described isn't exactly what is going on though. If you pass a list of length one, you will get an int (or 1d array) back. So sometimes lists will give ints and other times lists will give lists depending on their lengths which can cause problems.

samuelgarcia:
I agree with Cole with a list of size 1 is given then a list of size 1 should returned.

colehurwitz:
Ok, things are fixed now @alejoe91 . Are you ok with channel_locations and channel_groups returning as lists from now on @alejoe91 @samuelgarcia ?

PR #391, opened 2020-05-13
Caching - save_path arg and move_to function

alejoe91:


samuelgarcia:
Excelent.
Don't you want to keep an alias to old function to not break doozen of codes ;)

alejoe91:
No it's fine :)

PR #393, opened 2020-05-14
Wave_clus extractor using MATSortingExtractor

ferchaure:
Just a basic extractor for wave_clus.  I didn't add it to the extractors list

### DIFF
--- a/spikeextractors/cacheextractors.py
+++ b/spikeextractors/cacheextractors.py
@@ -10,13 +10,22 @@ import shutil
 
 
 class CacheRecordingExtractor(BinDatRecordingExtractor, RecordingExtractor):
-    def __init__(self, recording, chunk_size=None):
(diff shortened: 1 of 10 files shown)


alejoe91:
HI Fernando,

Could you call it `WaveClusSortingExtractor`? :)
Also please add it to the extractor list!

When this is merged, could you also modify the `spikesorters` wrapper? 
I guess that [this matlab part](https://github.com/SpikeInterface/spikesorters/blob/master/spikesorters/waveclus/p_waveclus.m#L61-L71) can be omitted, and the [get_result_from_folder](https://github.com/SpikeInterface/spikesorters/blob/master/spikesorters/waveclus/waveclus.py#L191-L202) could use the new extractor instead.

ferchaure:
you are right, I will look at that changes

alejoe91:
Thank @ferchaure 

Merging :)

PR #397, opened 2020-06-03
use read mode when opening nwb file

magland:
Is there a reason that nwb files were opened using 'a' (append) mode? It's causing a problem for me because it updates the create time of the file every time it is opened.


alejoe91:
HI @magland 

We can ask @bendichter and @luiztauffer !

alejoe91:
I think the reason it's because you can add metadata/processed data in a new session

bendichter:
@magland thanks for the fix. This looks good.

luiztauffer:
Hi @magland @alejoe91 
I can’t recall right now a situation where we needed to extract from an existing nwb file using `a` necessarily. So I suggest we go ahead and merge it!

PR #401, opened 2020-06-04
fix get traces by index

### LINKED ISSUE #400: Bug in NWB extractor - indeces for electrode table region are not channel_ids
In the code below the NWBExtractor uses the channel_ids to look up the indeces into the electrode table region of the e-series es. This produces an error for me where my electrode list has gaps (e.g. 1,2,3,5) but the electrodes.data[:] does not, as it indexes into the electrode table, which consists of consecutive entries.

The channel_ids here seem to refer to electrode_ids, which means that the code should be looking them up in nwbfile.electrodes to get their index in that table and then looking for that index in es.electrodes.data[:]

```
/Src/NWB/spikeextractors/spikeextractors/extractors/nwbextractors/nwbextractors.py in get_traces(self, channel_ids, start_frame, end_frame)
    257             nwbfile = io.read()
    258             es = nwbfile.acquisition[self._electrical_series_name]
--> 259             table_ids = [list(es.electrodes.data[:]).index(id) for id in channel_ids]
    260             if np.array(channel_ids).size > 1 and np.any(np.diff(channel_ids) < 0):
    261                 sorted_idx = np.argsort(table_ids)
```

luiztauffer:
fix #400


luiztauffer:
@alejoe91 I think this will solve it!

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -256,7 +256,7 @@ def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
         with NWBHDF5IO(self._path, 'r') as io:
             nwbfile = io.read()
             es = nwbfile.acquisition[self._electrical_series_name]
-            table_ids = [list(es.electrodes.data[:]).index(id) for id in channel_ids]
+            table_ids = [list(nwbfile.electrodes.id[:]).index(id) for id in channel_ids]
```
bendichter, line 259:
```suggestion
            table_ids = [es.electrodes[:].index.tolist().index(chan) for chan in channel_ids]
```

The above code will work, but assumes that the ElectricalSeries is referencing all of the electrodes in the electrodes able. That will work for NWB files created by SpikeExtractor, but won't work in the general case, which may include `ElectricalSeries` that correspond to specific subsets of electrodes. This suggestion bases the indexing on the ids of the electrodes that are in the subregion of the electrodes table that is referenced by the target `ElectricalSeries`, so it should be more robust

magland: I'm trying this out and the line @bendichter suggests seems to work, but causes massive computation delays -- like 60 seconds delay -- when number of channels is 192 and duration is 115 minutes.

Whereas I confirm that the solution in the PR works but has no such delay. But as Ben points out, maybe that isn't sufficient for all cases.

bendichter: @magland thanks for the report. I know what's going on and I'm pretty sure I can fix it.

bendichter: @magland I have edited the suggestion. Can you try that and confirm that it works and does not cause computational delays for you?

magland: @bendichter - in your edit, where is the `ephys_ts` variable coming from? I tried to use `nwbfile` in place but got an error: 'DynamicTable' object has no attribute 'table'

magland: Oh I used `es` and that seems to have worked. So I have:
```
            es_channel_ids = np.array(es.electrodes.table.id[:])[es.electrodes.data[:]].tolist()
            table_ids = [es_channel_ids.index(id) for id in channel_ids]
```
And I confirm that it works, and is efficient. Thx @bendichter!

alejoe91: @luiztauffer can you incorporate Ben's suggestion?

alejoe91:
Thanks @bendichter 

This should be ready to merge, right?

bendichter:
yup, thanks everyone!

PR #403, opened 2020-06-10
adds get_unsorted_spike_train method to waveclussortingextractor

ferchaure:


samuelgarcia:
Hi Fernando.
This method have to implemented also for the base class with a return None (maybe) when it is not implemented.
```
def get_unsorted_spike_train(self):
    return None
```

ferchaure:
Which is better, return None or raise a NotImplementedError on the baseextractor?

alejoe91:
Maybe a NotImplementedError? Or an empty np.array.. @samuelgarcia what do you think?

samuelgarcia:
Maybe the NotImplementedError is the wiser.

samuelgarcia:
I think it is OK for me.
I will add the same for tridesclous soon.

spikeextractors/baseextractor.py
```diff
@@ -274,6 +274,9 @@ def load_extractor_from_pickle(pkl_file):
             extractor._features = d['features']
         return extractor
 
+    def get_unsorted_spike_train(self):
```
alejoe91, line 277: @ferchaure can you move this to the SortingExtractor class? This base extractor is a common class for recording and sorting extractors which handles allocating memory, dumping etc

ferchaure: ok, sorry
should I include a _start_frame_ and _end_frame_ inputs?

alejoe91: That would be perfect!

PR #406, opened 2020-06-14
fix wc get_unit_ids

ferchaure:
I just realized that to write down the sorting results, the units should be integers.


colehurwitz:
Yep, that is correct. I can merge.

PR #412, opened 2020-06-18
fix read geom file for single channel

ferchaure:
I had to change the function to read the geometry file. 
 
A few lines after loading the file the number of channels on the recording is compared to self._geom.shape[0], but np.genfromtxt read a single line file as a vector of shape [2,] then it looks like the geom describes a 2 channel recording. 

The documentations say that loadtxt is equivalent to genfromtxt.


colehurwitz:
I think this makes sense. Should we merge @alejoe91 ?

alejoe91:
Looks good to me too :)

PR #421, opened 2020-07-07
Added NeuroscopeRecordingExtractor and minor additions, bug fixes to NeuroscopeSortingExtractor.

CodyCBakerPhD:
While it could possibly be a thin inherited class of the BinDatRecordingExtractor, the current version simply relies on the read_binary command for lazy data extraction and functions well for the desired use.

Also added some bug fixes for the NeuroscopeSortingExtractor which would occur if no IDs of '1' were present in the .clu file, but those of higher index were.

Also introduced a new function to the NeuroscopeSortingExtractor that forms a list of such objects read over multiple shanks; this list would then be formed by concatenate_sorting into a single MultiSortingExtractor object.


CodyCBakerPhD:
The current commit is having travis-ci trouble with a bs4 object, but everything runs fine when I try it locally. Any ideas?

alejoe91:
You probably need to add the packages in the . travis.yaml

setup.py
```diff
@@ -18,7 +18,7 @@
     packages=setuptools.find_packages(),
     package_data={},
     install_requires=[
-        'numpy',
+        'numpy','bs4'
     ],
```
samuelgarcia, line 22: Hi.
The idea in spikeextractors is that every new extractor have only weak dependency.
In short, we have in setup.py very few dep.

The method we use is something like that

```python
try:
    import bs4
    HAVE_BS4 = True
except ImportError:
    HAVE_BS4 = False


class NeuroscopeRecordingExtractor(RecordingExtractor):
    def __init__(self):
        assert HAVE_BS4, 'please install bs4 to use this extractor'
```



```

File: spikeextractors/extractors/openephysextractors/openephysextractors.py [hunk 2]
```diff
@@ -19,9 +19,9 @@ class OpenEphysRecordingExtractor(RecordingExtractor):
     mode = 'folder'
     installation_mesg = "To use the OpenEphys extractor, install pyopenephys: \n\n pip install pyopenephys\n\n"  # error message when not installed
 
-    def __init__(self, folder_path, *, experiment_id=0, recording_id=0, dtype='float'):
+    def __init__(self, folder_path, *, experiment_id=0, dtype='int32'):
         assert HAVE_OE, self.installation_mesg
-        assert dtype == 'int16' or 'float' in dtype, "'dtype' can be int16 (memory map) or 'float' (load into memory)"
+        assert dtype == 'int16' or 'int32' in dtype, "'dtype' can be int16 or 'int32' (both memory map)"
```
alejoe91, line 24: The pyopenephys package only supports float and int16. The Open Ephys hardware records int16 (that can be memmapped). However, to get uV, one needs to multiply by the gain, which is done at the `get_traces` level. 
Please revert this change

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,330 @@
...
+        Parameters
+        ----------
+        unit_id: int
+            The unit_id of the unit to be added.
+        '''
+        self._unit_ids.append(unit_id)
+        self._spiketrains.append(spike_times)
+        
+        
+    def read_shanks(general_path, shanks, keep_mua_units=True):
```
alejoe91, line 239: I think we could do this systematically at the constructor level. If `res_file` and `clu_file` are used as args, the constructor can laod a single file. If instead a `folder_path` is used, we can loop through the folder, find all shanks files (suffix .res.*/.clu.*) and load them in sequence. Moreover, we can have an extra argument in the constructor, `exclude_shanks` which is `None` by default, but if an int or a list of ints are passed we can exclude those shanks from loading. Sounds good?

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,330 @@
+from spikeextractors import RecordingExtractor,SortingExtractor
+from spikeextractors.extractors.bindatrecordingextractor import BinDatRecordingExtractor
+import numpy as np
+from pathlib import Path
+from spikeextractors.extraction_tools import check_get_traces_args,check_valid_unit_id,read_binary
+from bs4 import BeautifulSoup
+import os
+
+class NeuroscopeRecordingExtractor(RecordingExtractor):
```
alejoe91, line 9: I think we can change that and inherit from the BinDatRecordingExtractor later

alejoe91:
> The current commit is having travis-ci trouble with a bs4 object, but everything runs fine when I try it locally. Any ideas?

From the `bs4` documentation: https://www.crummy.com/software/BeautifulSoup/bs4/doc/#quick-start

To use the xml/lxml parser, you need to install the lxml python package: `pip install lxml`
Adding this to the `.travis` should fix the tests

setup.py
CodyCBakerPhD: Fixed!

spikeextractors/extractors/openephysextractors/openephysextractors.py
CodyCBakerPhD: Change reverted

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Still working on this, will let you know when it's done.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Copying the structure from the phy version worked well, passing tests on my end.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,306 @@
...
+        """ Convert and save the recording extractor to Neuroscope format
+
+        parameters
+        ----------
+        recording: RecordingExtractor
+            The recording extractor to be converted and saved
+        save_path: str
+            Full path to desired target folder
+        """
+        RECORDING_NAME = save_path + '.dat'
```
alejoe91, line 74: Appending `.dat` is not neede as it's already handled by the `BinDatRecordingExtractor.write_recording()`, which calls the [`write_to_binary_dat_format`](https://github.com/SpikeInterface/spikeextractors/blob/f0d2653f6e55788150d0fcba4242aee5c75b8988/spikeextractors/extraction_tools.py#L294-L298) function.

So you can remove lines 74 and 78 and just pass `save_path` directly to the `write_recording`

CodyCBakerPhD: Got it. Also, it seems as if I'm unable to personally alter the travis.yaml (see conflicting files) to include bs4 and lxml, is that something you have to do?

alejoe91: So you should add this repo as a remote upstream locally:

`git remote add upstream https://github.com/SpikeInterface/spikeextractors.git`

Then you can locally run (after you have committed your changes):

`git merge upstream/master`

Some conflicts will be there, you can resolve them locally, re-commit, and push. Then they should be fixed :)

CodyCBakerPhD: The remote upstream is already setup, and claims to be up to date on merge attempt. The latest commit to the /catalystneuro/spikeextractor fork, NeuroscopeRecordingExtractor branch, directly altered the travis.yaml at that location to include bs4 and lxml. After that commit, the travis-ci won't run because it shows the file conflicting.

alejoe91: But do you have conflicts locally? If so, you have to resolve them, add the files, recommit, and push. Usually I'm able to fix it from github, but it doesn't seem possible now. If that doesn't work I can try to push on the branch directly later.

CodyCBakerPhD: No conflicts locally, the message is the one at the bottom of the pull request page where you usually see the travis results.

CodyCBakerPhD: And that "Only those with write access to this repository can merge pull requests."

alejoe91:
It doesn't let me merge either..strange! Can you try to make another branch and make a new PR? Let's see if the problem is still there

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
alejoe91: @CodyCBakerPhD when I checkout youor fork and run `git merge upstream/master` I got a merging conflict to the `.travis.yml`. Unfortunately I don't have write permissions on your fork. Next time you make a PR, there should be a checkbox sayng something like: allow owner to make changes. If you check that box, then I would be able to push on the repo. I think this is also why I can't resolve the conflict here: I can't commit and push to your fork.

alejoe91:
Did you have to pull upstream/master? :)

.travis.yml
```diff
@@ -6,8 +6,7 @@ services:
 python:
 - '3.6'
 install:
-# - pip install shybrid
-- pip install mountainlab_pytools pynwb==1.3.3 pyopenephys kbucket h5py MEArec==1.4.1 exdir ruamel.yaml nixio neo>=0.8 hdf5storage
```
alejoe91, line 10: You need to comment our the shybrid install and point MEArec to 1.4.1, as it was in master

CodyCBakerPhD:
No, I went in via the GitHub PR page near the 'write access' button where it shows the conflicts in the merging; the code portion of that is interactable, so you can manually type in what you want to add/delete from that screen and then OK the merge. Nothing I did on my local end or via git commands ever did anything. Trying to fix that other build issue now.

.travis.yml
CodyCBakerPhD: Done! Current version of the travis.yaml should be identical to the master on SpikeInterface, but with bs4 and lxml added.

CodyCBakerPhD:
I'll work more on the other debugging tomorrow, as well as integrating read_shank into __init__ for the SortingExtractor

alejoe91:
Note that also the recording extractor tests are failing due to the int16 int32 assertion. But that should be a quick fix!

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -245,17 +246,19 @@ def __init__(self, resfile_path=None, clufile_path=None, folder_path=None, keep_
...
                     # Maybe the issue lies in calling NeuroscopeSortingExtractor() within its own __init__?
-                    all_shanks_list_se.append(NeuroscopeSortingExtractor(resfile_path=resfile_path,
-                                                                         clufile_path=clufile_path,
-                                                                         keep_mua_units=keep_mua_units))
+                    # all_shanks_list_se.append(NeuroscopeSortingExtractor(resfile_path=resfile_path,
+                    #                                                      clufile_path=clufile_path,
+                    #                                                      keep_mua_units=keep_mua_units))
                     
                     
                 MultiSortingExtractor.__init__(self, all_shanks_list_se)
```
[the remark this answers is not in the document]
colehurwitz: Maybe you can make the `NeuroscopeSortingExtractor` inherit from the `MultiSortingExtractor` object? Then, if it auto-detects a single .res .clue pair, it can be a `MultiSortingExtractor` with a single `SortingExtractor` (which is fine). Otherwise, it can naturally be a `MultiSortingExtractor` with multiple `SortingExtractors`. Part of the challenge is that we currently don't have a mechanism for dealing with multi-segment recordings in the original `SortingExtractor`, which makes this kinda file forma difficult. Thoughts on this potential solution @alejoe91 @samuelgarcia ?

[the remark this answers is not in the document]
CodyCBakerPhD: To the best of my understanding, defining the entire `NeuroscopeSortingExtractor` as a subclass of `MultiSortingExtractor` leads to a chicken and egg problem, where the `MultiSortingExtractor` requires, as its `_sorting` internal or `sorting` argument, a distinct type of sorting extractor with its own defined functions such as `get_unit_ids `and `get_traces`.

Perhaps the solution would be to have a `NeuroscopeSortingExtractor(SortingExtractor)` for single shank data, and a `NeuroscopeMultiSortingExtractor(MultiSortingExtractor)` for multi-shank, which uses the `NeuroscopeSortingExtractor` as the filling sorting type?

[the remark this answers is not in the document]
colehurwitz: These seems like a good solution for now. We are thinking about supporting multiple segments (for multi-shank) in future versions, but that is still being discussed.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,387 @@
...
+                    
+                    # runs correct, but final object does not inherit basic functions like 'get_unit_ids' from MultiSortingExtractor
+                    # and also doesn't seem to set the interval values of _all_unit_ids and such correctly...
+                    # Maybe the issue lies in calling NeuroscopeSortingExtractor() within its own __init__?
+                    # all_shanks_list_se.append(NeuroscopeSortingExtractor(resfile_path=resfile_path,
+                    #                                                      clufile_path=clufile_path,
+                    #                                                      keep_mua_units=keep_mua_units))
+                    
+                    
+                MultiSortingExtractor.__init__(self, all_shanks_list_se)
```
alejoe91, line 264: I think that what you are missing is the following:

`self = MultiSortingExtractor.__init__(self, all_shanks_list_se)`

This basically resets the object to be a `MultiSortingExtractor`, hence inheriting all the `SortingExtractor` methods.

The same holds for line 235:
`self = NeuroscopeSortingExtractor.__init__(self, resfile_path=resfile_path,
                                                          clufile_path=clufile_path,
                                                          keep_mua_units=keep_mua_units)`

I think this could also be a viable solution in order not to have 2 separate extractors as @CodyCBakerPhD suggested

alejoe91: Moreover, @CodyCBakerPhD you sould set the group property for each unit: at line 252:

```
single_sorting = NeuroscopeSortingExtractor(resfile_path=resfile_path, clufile_path=clufile_path, 
                                                                              keep_mua_units=keep_mua_units)
for u in single_sorting.get_unit_ids():
    single_sorting.set_unit_property(u, 'group', int(shank_id))
all_shanks_list_se.append(single_sorting)
```

So when you create the `MultiSortingExtractor`, each unit carries info about its group :)

CodyCBakerPhD: In the end it doesn't seem to matter if I do `self = MultiSortingExtractor(sortings=all_shanks_list_se)`  or `MultiSortingExtractor.__init__(self,sortings=all_shanks_list_se)` to create and fill that object; the issue is apparently the dual overriding of the `get_unit_ids()` function, which calls different internal names for the two base types (`_unit_ids` or `_all_unit_ids`). If the entire NeuroscopeSortingExtractor class inherits from SortingExtractor, everything works during initialization, and for single shank conditions, but for multi-shank the `get_unit_ids()` calls the function specific to NeuroscopeSortingExtractors instead of the MultiSortingExtractor, even if we set `self = MultiSortingExtractor(sortings=all_shanks_list_se)`. Likewise, if we inherit from the MultiSortingExtractor and allow it to override that function definition, then the single shank versions stop working since they use the different internal naming conventions. 

Similar issues to this would also arise for `get_unit_spike_train()`. 

I can only see two fixes to this: 

1) Adapt the `get_unit_ids()` function for the NeuroscopeSortingExtractor class to identify whether or not it is a single shank or multi shank and call the correct corresponding id retrieval method. Ditto for spike trains.

2) Two separate classes, with the NeuroscopeMultiSortingExtractor using thin inheritance, and the NeuroscopeSortingExtractor being exactly what it is now without the multi-shank condition (which is all working as intended).

I'm inclined to go with solution 2 here since it keeps the application of the methods abstract and appropriate to their final use. This can then be refactored later if a better base approach to multiple data sources is incorporated into the extractors.

alejoe91: Hi @CodyCBakerPhD 

I though it would have worked! I also prefer solution 2 then. The user would now if there are multiple shanks or not, or at least this is easily discoverable by looking at the folder. 

Thanks for looking into this :)

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+    Extracts raw neural recordings from large binary .dat files in the neuroscope format.
+    
+    The recording extractor always returns channel IDs starting from 0.
+    
+    The recording data will always be returned in the shape of (num_channels,num_frames).
+
+    Parameters
+    ----------
+    folder_path : str
+        Path to the folder containing the .dat file.
```
alejoe91, line 31: Can you add the `subset_channels` to the docstring? Actually I would call it `channel_ids`, which is how we call it elsewhere

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+        parameters
+        ----------
+        recording: RecordingExtractor
+            The recording extractor to be converted and saved
+        save_path: str
+            Full path to desired target folder
+        dtype: str
+            Data type to be used in writing. Will throw a warning if stored recording type from get_traces() does not match.
+        """
+        _, RECORDING_NAME = os.path.split(save_path)
```
alejoe91, line 84: would this fail if one passes a relative path? e.g. `save_path='recording'`

I recommend looking into `pathlib` for handling paths, but let's keep it for later. I can quickly modify it once it's merged

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+            soup = BeautifulSoup("",'xml')
+
+            new_tag = soup.new_tag('nbits')
+            dtype = recording.get_dtype()
+            
+            if not any([dtype == x for x in ['int16', 'int32']]):
+                print('Warning: Unsupported data type passed (',dtype,'); converting to default of int32!',sep="")
+                dtype = 'int32'
+                
+            n_bits = str(dtype)[3:5]
```
alejoe91, line 101: This will fail if one passes a `numpy` dtype. I see in line 97 that you only accept strings, so it's fine here. In general, a more robust way is to find the `int` substring and use relative indexes from that position :)

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+    keep_mua_units : bool
+        Optional. Whether or not to return sorted spikes from multi-unit activity. Defaults to True.
+    """
+    extractor_name = 'NeuroscopeSortingExtractor'
+    installed = True  # check at class level if installed or not
+    is_writable = True
+    mode = 'custom'
+    installation_mesg = ""  # error message when not installed
+
+    def __init__(self, resfile_path=None, clufile_path=None, folder_path=None, keep_mua_units=True):
```
alejoe91, line 157: Now with the `NeuroscopeMultiSortingExtractor` the `folder_path` arg is not needed here anymore, right? same for all the logic using it

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+                    for s_id in self._unit_ids:
+                        self._spiketrains.append(res[(clu == s_id).nonzero()])
+                else:
+                    # Ignoring IDs of 0 until get_unsorted_spike_train is implemented into base
+                    # Also ignoring IDs of 1 since user called keep_mua_units=False
+                    n_clu -= 2;
+                    self._unit_ids = [x+1 for x in range(n_clu)] # generates list from 1,...,clu[0]-2
+                    for s_id in self._unit_ids:
+                        self._spiketrains.append(res[(clu == s_id+1).nonzero()]) # only reading cluster IDs 2,...,clu[0]-1
+        elif resfile_path is None and clufile_path is None:
```
alejoe91, line 214: This should be removed

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+        Optional. Path to the collection of .res and .clu text files. Will auto-detect format.
+    keep_mua_units : bool
+        Optional. Whether or not to return sorted spikes from multi-unit activity. Defaults to True.
+    exclude_inds : list
+        Optional. List of indices to ignore. The set of all possible indices is chosen by default, extracted as the final integer of all the .res.%i and .clu.%i pairs.
+    """
+    extractor_name = 'NeuroscopeMultiSortingExtractor'
+    installed = True  # check at class level if installed or not
+    is_writable = True
+    mode = 'custom'
```
alejoe91, line 356: this mode should be `folder` (accepted modes are `folder`, `file`, `custom`)

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+    
+    The recording data will always be returned in the shape of (num_channels,num_frames).
+
+    Parameters
+    ----------
+    folder_path : str
+        Path to the folder containing the .dat file.
+    """
+    extractor_name = 'NeuroscopeRecordingExtractor'
+    installed = True  # check at class level if installed or not
```
alejoe91, line 34: `installed = HAVE_BS4_LXML` 

same below

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -0,0 +1,451 @@
...
+        # create parameters file if none exists
+        if not os.path.isfile(save_xml):
+            soup = BeautifulSoup("",'xml')
+
+            new_tag = soup.new_tag('samplingrate')
+            new_tag.string = str(sorting.get_sampling_frequency())
+            soup.append(new_tag)
+
+            # write parameters file
+            f = open(save_xml, "w")
```
alejoe91, line 449: generally, for files I would use context to make sure that the file is properly closed (even if you forget the close())

```
with open(save_xml, "w") as f:
    f.write(str(soup))
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -418,8 +418,8 @@ def add_electrical_series(recording, nwbfile, metadata):
         """
         # ElectricalSeries
         if 'ElectricalSeries' not in metadata['Ecephys']:
-            metadata['Ecephys']['ElectricalSeries'] = [{'name': 'ElectricalSeries',
-                                                        'description': 'electrical_series_description'}]
+                    metadata['Ecephys']['ElectricalSeries'] = [{'name': 'ElectricalSeries',
```
alejoe91, line 421: is this extra indentation correct?

spikeextractors/extractors/openephysextractors/openephysextractors.py
```diff
@@ -19,7 +19,7 @@ class OpenEphysRecordingExtractor(RecordingExtractor):
     mode = 'folder'
     installation_mesg = "To use the OpenEphys extractor, install pyopenephys: \n\n pip install pyopenephys\n\n"  # error message when not installed
 
-    def __init__(self, folder_path, *, experiment_id=0, recording_id=0, dtype='float'):
+    def __init__(self, folder_path, *, experiment_id=0, dtype='int32'):
```
alejoe91, line 22: can your revert this? 
An open ephys file can include multiple experiments and multiple recordings (so we need `recording_id`). I'd rather have `float` by default, so that uV are returned.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Yes, relative paths could definitely cause issues if the passed path was not in correspondence to the working directory at the time of the call. That said, we only ever pass absolute paths in the functions that call these extractors, so I'll just add that requirement for now in the docstring.

For our purposes, I can't see needing the ability to do a better job on relative paths since we never use these in a scripting manner that would tend to rely more heavily on working directory location; but if you think there is potential for other users that may want to use the functions in such a way, I leave that to your discretion.

CodyCBakerPhD: More importantly, I suppose, is that the RECORDING_NAME may not have to be exactly the same as the last element of the save_path, though it simply happens to be in the datasets we are prototyping on. I'll try to generalize that a bit more as well.

CodyCBakerPhD: Actually, never mind. It's pretty easy to just coerce the absolute path and extract the specific name rather than expecting adherence to any particular format or directory. Fixing this now.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Even for the NeuroscopeSortingExtractor, the new folder_path option still allows auto-detection of the .res and .clu files at the location. I've also added the tests for that functionality. It's still back-compatible to the old format as well.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: What does this specify? If the NeuroscopeRecordingExtractor is adjusted to take a file path instead of a folder path, should I change that to mode = 'file' as well?

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: You mean the final 'else' statement at the end of that switch, because the previous assertions mean it would never get called?

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Fixed!

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Fixed in all cases! That is indeed good practice, I'll try to get into that habit.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Disregard. I've adjusted the majority of that logic to be more coherent (hopefully).

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: So, apparently even the BinDatRecordingExtractor doesn't actually subset by `channel_ids`, it simply reads the first `len(channel_ids)` channels of that binary format. I've adjusted the variable name to more accurately reflect this and added it to the docstring!

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Great point; I've implemented the robust way now!

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: `exclude_shanks `is now incorporated into the `NeuroscopeMultiSortingExtractor`

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: I've tried to generalize this as much as possible. The paths that are passed within the test_extractor.py need a particular format to work with the absolute conversions, but theoretically, a user could pass a really sparse relative path and it _should_ work.

spikeextractors/extractors/openephysextractors/openephysextractors.py
CodyCBakerPhD: Fixed! My sincerest apologies, I'm not sure why these changes to `openephys `were made on my end.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Fixed! Copy/paste issue.

CodyCBakerPhD:
The test_extractor is completing without error on my end, so will take me some time over the weekend to figure out what travis doesn't like about the file name structure.

alejoe91:
> The test_extractor is completing without error on my end, so will take me some time over the weekend to figure out what travis doesn't like about the file name structure.

You used \ to build the xml path. This could work for windows, but it'd definitely fail with Unix systems. You should use os.path or parking to handle any path

alejoe91:
@CodyCBakerPhD I made some changes:

- improve the use of `pathlib` 
- simplify the logic in many places
- NeuroscopeMultiSortingExtractor can also handle SortingExtractor objects with mutiple groups

I have one last doubt: if the `.xml` exists, when writing the recorsing/sorting, should we raise a `FileExistsError` or remove and overwrite?

CodyCBakerPhD:
I would raise an error; the writing methods are really only there for the testing framework and we won't be using them at all ourselves. The actual .XML that are in the databases we're converting have far richer information in them than what we use here, so we would never want to overwrite those.

alejoe91:
Sounds good! The other changes look good for you?

CodyCBakerPhD:
The use of `pathlib `is great, and really cleaned up some of that logic! Two quick things:

1) It looks like you removed the `numchan `argument from `NeuroscopeRecordingExtractor`? It probably seems odd but we do actually need it for what we're doing on the Buzsaki dataset; the number of channels in the .XML file doesn't actually correspond to the number of channels we want to extract.

2) Did you want to add anything in the docstring for the `NeuroscopeMultiSortingExtractor `to reflect the bit about the `group` unit properties? We probably won't make use of this for our purposes since that information is being set in a more general dictionary of unit metadata stored as a unit property for the NWB writing procedure.

alejoe91:
> The use of `pathlib `is great, and really cleaned up some of that logic! Two quick things:
> 

I know!!! :)

> 1. It looks like you removed the `numchan `argument from `NeuroscopeRecordingExtractor`? It probably seems odd but we do actually need it for what we're doing on the Buzsaki dataset; the number of channels in the .XML file doesn't actually correspond to the number of channels we want to extract.

Yes, mainly because it forces you to load the first nchannels. Do you mean that in the xml you might have num channels=32 but the actual recording is 64 channels? In case you want to extract a subset of channels, it can be done later very easily with the `SubRecordingExtractor` (and that also allows you to choos the channel_ids). Would that work for the Buszaki dataset?

> 2. Did you want to add anything in the docstring for the `NeuroscopeMultiSortingExtractor `to reflect the bit about the `group` unit properties? We probably won't make use of this for our purposes since that information is being set in a more general dictionary of unit metadata stored as a unit property for the NWB writing procedure.

Yes, I think we could add something like: "The group id is loaded as unit property 'group'"

CodyCBakerPhD:
Yeah, what happens in the data is there are actually 80 channels, but only the **first** 64 are the shank channels we want, so it works to just load in the first 64, though more generally we would definitely want to pass the channel ids and subset from that. I'll try to get a push in the next day using the `SubRecordingExtractor ` for this; so long as the way it works doesn't actually access the data in the memmap (i.e., `data[channel_ids,:]`).

alejoe91:
> Yeah, what happens in the data is there are actually 80 channels, but only the **first** 64 are the shank channels we want, so it works to just load in the first 64, though more generally we would definitely want to pass the channel ids and subset from that. I'll try to get a push in the next day using the `SubRecordingExtractor ` for this; so long as the way it works doesn't actually access the data in the memmap (i.e., `data[channel_ids,:]`).

Are the other channels still neural data? or are they like ADC data for behavior or similar? In case they are neural data, I think we should go for the SubRecordingExtractor, otherwise I think it's ok to remove them.

CodyCBakerPhD:
They are special environmental electrodes recorded inline with the neural data; we only include the processed lfp for these channels.

CodyCBakerPhD:
In the case that you're OK with having those extra channels removed by the `numchan `approach, can you revert that change? Then we should be good for what we need right now; can always extend it later if needed.

alejoe91:
> In the case that you're OK with having those extra channels removed by the `numchan `approach, can you revert that change? Then we should be good for what we need right now; can always extend it later if needed.

I think it's better to use the `SubRecordingExtractor`, since it is a specific case for these datasets. It's only an extra line of code:
`rec = se.SubRecordingExtractor(rec, channel_ids=np.arange(numchan)`

CodyCBakerPhD:
OK, we should be able to use the current extractors plus that extra line of `SubRecordingExtractor` to do what we need. Everything else looks good!

PR #429, opened 2020-07-21
Preserve memmap when setting features

alejoe91:
Reordering the indexes and the values converts memmap to arrays, which is not going to work for large files.
Removed it


alejoe91:
@colehurwitz I ran into this computing waveforms with memmap and getting gb of RAM usage :)

PR #430, opened 2020-07-27
NWBSortingExtractor write_sorting extended functionality

CodyCBakerPhD:
Extended the NWBSortingExtractor write_sorting method to have greater capability in handling custom data columns for spiking units, encoded as unit_properties within the sorting extractor, specifically as a dictionary of the form `{'name': my_name, 'description': my_description, 'data': my_data}` that is also the form used by pynwb functions to add the custom data. This allows the user to pass meaningful descriptions along with the actual metadata corresponding to some experiment.

Also added functionality for writing MultiSortingExtractor objects if passed; this merely extracts the relevant properties from each object in the `.sorting` list. It also fills a custom column data of the `shank id` in the writing procedure.

Also included a couple of random bug fixes for detecting the existence of higher-level parts to a dictionary in the NeuroscopeRecordingExtractor add_electrical_series, which will create the necessary key if it does not exist.


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -685,27 +689,38 @@ def frame_to_time(self, frame):
...
         fs = sorting.get_sampling_frequency()
-        if hasattr(sorting, '_t0'):
-            t0 = sorting._t0
-        else:
-            t0 = 0.
 
+        if isinstance(sorting,se.MultiSortingExtractor):
+            total_prop_dict = {}
+            counter = 0
+            for this_sorting in sorting.sortings:
```
alejoe91, line 712: for this you could also use `for i, this_sorting in enumerate(sorting.sortings):`
`i` would be your counter automatically

samuelgarcia:
@alejoe91 : if the ".dat" is in the git history could you use the option "squash and merge" after the file removal because it stays in git history and make the git clone quite heavy.

alejoe91:
@samuelgarcia how do you do that? Maybe better to make a new PR

samuelgarcia:
No on the "merge button" (green) there is now an option with the context menu.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Done. Where does the cache_rec.dat come from? I added it to the git .ignore as well.

alejoe91: I think it gets created by the tests. Maybe was added if you did git add * ..is it possible?

CodyCBakerPhD: Oh, it was definitely added by git add *, I just have no idea where the file came from, what it contained, or why it exists. Shouldn't the testing scripts clean up stuff like that?

alejoe91: It should, but I recommend adding the files staged for commit only or adding them manually. That file is not staged for commit. Instead of git add * you can do git commit -am to add only staged files.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -727,51 +740,97 @@ def write_sorting(sorting, save_path, **nwbfile_kwargs):
                     spkt = sorting.get_unit_spike_train(unit_id=id) / fs # spike trains withinin the SortingExtractor object are not scaled by sampling frequency
                     nwbfile.add_unit(id=id, spike_times=spkt)
 
+            # Always add column for shank ids if a MultiSortingExtractor
+            if isinstance(sorting,se.MultiSortingExtractor):
+                set_dynamic_table_property(
+                        dynamic_table=nwbfile.units,
+                        row_ids=[int(k) for k, v in sorting._unit_map.items()],
+                        property_name='shank_id',
```
alejoe91, line 748: I would call this `group`, which is more general. For tetrodes, for example, you have multiple groups, but they are not on different shanks.

alejoe91: I'm fixing this real quick :)

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -727,51 +740,97 @@ def write_sorting(sorting, save_path, **nwbfile_kwargs):
...
+
             # Units properties
             for pr in all_properties:
-                unit_ids = [int(k) for k, v in sorting._properties.items()
+                unit_ids = [int(k) for k, v in total_prop_dict.items()
                             if pr in v]
-                vals = [v[pr] for k, v in sorting._properties.items()
-                        if pr in v]
+                
+                # We can always guarantee two levelx of nested dictionaries based
```
alejoe91, line 759: This is unclear.. what do you mean?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -727,51 +740,97 @@ def write_sorting(sorting, save_path, **nwbfile_kwargs):
...
+                # based on the current structure
+                # If the value inside that is also a dictionary, require that 
+                # it is of the {'name': my_name, 'description':my_description, 'data':my_data} format
```
alejoe91, line 762: What is this dictionary?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -727,51 +740,97 @@ def write_sorting(sorting, save_path, **nwbfile_kwargs):
...
+                        vals = [v[pr]['data'] for k, v in total_prop_dict.items()]
+                else:
+                    name = pr
+                    vals = [v[pr] for k, v in total_prop_dict.items()
+                            if pr in v]
+                    description = 'no description'
+                
+                # Special case of setting max_electrodes requires a table to be
+                # passed to become a dynamic table region
+                if name == 'max_electrode':
```
alejoe91, line 789: this is `max_channel` in SI usually

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Sorry, it should read

# We can always guarantee two levels of nested dictionaries
# based on the current structure

which refers to the fact that if there is a unit property, it was added via the basic set_unit_property, which by line 331 of `SortingExtractor`

`self._properties[unit_id][property_name] = value`

implies there are two levels to the dictionary such that 

`unit_prop = next(iter(next(iter(total_prop_dict.values())).values()))`

will not return an error by trying to iterate or call values over a non-dictionary object.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: As I said in the top comment, it is the pynwb dictionary format for adding custom columns to the units after the units have been added.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Can you provide a source/citation for this? It is current referred to as `max_electrode` on the current conversions (and formats currently on DANDI), so we would want to ensure that convention is correct.

alejoe91:
@CodyCBakerPhD 
- I fixed the issue of missing properties for some units by setting the value to `undefined`. 
- When saving and loading `max_channel` or `max_electrode`, even if it's saved as a reference to the ElectrodeTable, when using the `NwbSortingExtractor` the index of the pandas dataframe is used as property

alejoe91:
@CodyCBakerPhD @bendichter this is good to go IMO.

I had to make some changes to be able to save to Nwb missing properties for certain units (now saved as `nan`) and not to load them back in SI.

CodyCBakerPhD:
@alejoe91 Looks good, thanks for patching those last couple issues. Tested it out and it's working correctly! I can't attest to the changes to the actual SortingExtractor __init__ using the panda frame, since I only use the writing function on other types of SortingExtractors.

alejoe91:
@CodyCBakerPhD I tested it by saving.a `max_electrode` property and then load it back. Seems to work :)

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -416,7 +420,9 @@ def add_electrical_series(recording, nwbfile, metadata):
         Auxiliary static method for nwbextractor.
         Adds traces from recording object as ElectricalSeries to nwbfile object.
         """
-        # ElectricalSeries
+        if 'Ecephys' not in metadata:
```
bendichter, line 423: I think the structure of this metadata dict is going to change

CodyCBakerPhD: This particular change is mostly an error-check if the user ever calls NwbRecordingExtractors.add_electrical_series directly in isolation. If called from the NwbRecordingExtractor.write_recording, the base keys are already created by the surrounding add functions.

Do you mean the passed metadata dictionary is soon to require revamping to adhere to the json schema from the new converter format? If so, then yeah we'll have to go through some of these helper functions to adjust the defaults.

bendichter: > Do you mean the passed metadata dictionary is soon to require revamping to adhere to the json schema from the new converter format?
Yes, that's what I mean

alejoe91: So should we modify this or can we merge as it is? We can always modify it later :)

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -633,7 +639,12 @@ def __init__(self, file_path, electrical_series=None, sampling_frequency=None):
                 else:  # if it is unit_property
                     for id in units_ids:
                         ind = list(units_ids).index(id)
-                        self.set_unit_property(id, item, nwbfile.units[item][ind])
+                        if isinstance(nwbfile.units[item][ind], pd.DataFrame):
```
bendichter, line 642: @alejoe91 in what situation would `nwbfile.units[item][ind]` be a `pd.DataFrame`? Is this something you added for tests?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -724,54 +744,110 @@ def write_sorting(sorting, save_path, **nwbfile_kwargs):
...
+                        name = names[0]
+                        description = descriptions[0]
+                        for u in all_units:
+                            if pr in sorting.get_unit_property_names(u):
+                                vals.append(sorting.get_unit_property(u, pr)['data'])
+                            else:
+                                vals.append(np.nan)
+                    else:
+                        print('Warning: NwbSortingExtract.write_sorting() requires, if a unit property' +
+                              'is a dictionary, the keys must be "name, description, data" in order!')
```
bendichter, line 793: why "in order"?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -724,54 +744,110 @@ def write_sorting(sorting, save_path, **nwbfile_kwargs):
...
+                for u in all_units:
+                    if pr in sorting.get_unit_property_names(u):
+                        unit_prop = sorting.get_unit_property(u, pr)
+                        break
+                # unit_prop = next(iter(next(iter(total_prop_dict.values())).values()))
+                print('Unit prop', pr, unit_prop)
+                if type(unit_prop) is dict:
+                    item_names = [x for x in unit_prop.keys()]
+                    # This checks for the ordering as well
+                    if sorted(item_names) == sorted(['name', 'description', 'data']):
```
bendichter, line 770:
```suggestion
                    if set(item_names) == set(['name', 'description', 'data']):
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: it should not be in order

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: @CodyCBakerPhD added a `table` arg to the [`set_dynamic_table_property ` ](https://github.com/SpikeInterface/spikeextractors/pull/430/files#diff-35851cf23629fc4a9bddde08ed99b1c2R31).

When the sorting have `max_channel` or `max_electrode` properties, the [`nwbfile.electrodes`](https://github.com/SpikeInterface/spikeextractors/pull/430/files#diff-35851cf23629fc4a9bddde08ed99b1c2R807) are used as `table` for the `add_column` function. 

When retrieving those properties, they are returned as a pandas df with electrode id as index

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: yeah it could be in any order, it's a dictionart

CodyCBakerPhD:
@alejoe91 Pretty big refactor of the `write_sorting` method per our discussion. All that unit property dictionary stuff is replaced by the new `property_descriptions` argument and the references to internal `_.property` and `._features` in the early-mid parts of the function are replaced by the more appropriate access conventions. Also added the capability to pass the `nwbfile` object as an argument in both `write_sorting` and `write_recording`. 

Removing it from the last part about writing spike features would take a bit more work and I'm inclined to include that when we refactor the waveform stuff so we have ways to functionally test it on actual data.

Wasn't sure how you would want to go about adding tests for the `nwbfile` argument in both of those, so I left that as a `#TODO`.

@bendichter These probably look like pretty hefty changes based on the amount of green in the change view, but 95% of that is stuff Alessio requested unrelated to the property descriptions.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -416,7 +410,9 @@ def add_electrical_series(recording, nwbfile, metadata):
         Auxiliary static method for nwbextractor.
         Adds traces from recording object as ElectricalSeries to nwbfile object.
         """
-        # ElectricalSeries
+        if 'Ecephys' not in metadata:
+            metadata['Ecephys'] = {}
+
         if 'ElectricalSeries' not in metadata['Ecephys']:
```
bendichter, line 416:
```suggestion
        if 'Ecephys' not in metadata or 'ElectricalSeries' not in metadata['Ecephys']:
```

this can be used in place of `metadata['Ecephys'] = {}`

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -501,81 +497,111 @@ def add_epochs(recording, nwbfile):
...
+                read_mode = 'w'
+    
+            with NWBHDF5IO(save_path, mode=read_mode) as io:
+                if read_mode == 'r+':
+                    nwbfile = io.read()
+                else:
+                    if 'NWBFile' not in metadata:
+                        metadata['NWBFile'] = {'session_description': 'no description',
+                                               'identifier': str(uuid.uuid4()),
+                                               'session_start_time': datetime.now()}
```
bendichter, line 587: this should use `update_dict` to handle if metadata['NWBFile'] is partially defined

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -681,68 +712,68 @@ def time_to_frame(self, time):
...
+        all_properties = set()
+        all_features = set()
+        for unit_id in unit_ids:
+            all_properties.update(sorting.get_unit_property_names(unit_id))
+            all_features.update(sorting.get_unit_spike_feature_names(unit_id))
+        
+        if property_descriptions is not None:
+            property_description_names = set(property_descriptions.keys())
+            print(all_properties)
+            print(property_description_names)
```
bendichter, line 735: keep these print statements?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -681,68 +712,68 @@ def time_to_frame(self, time):
...
-                    default_value=np.nan,
-                    description='no description'
-                )
-
+                property_descriptions.update({pr: 'no description'})
+        
+        # If no Units present in nwb file
+        if nwbfile.units is None:
+            for pr in all_properties:
+                # Special case of setting max_electrodes requires a table to be
```
bendichter, line 749:
```suggestion
                desc = property_descriptions.get(pr, 'no description')
                # Special case of setting max_electrodes requires a table to be
```

This should obviate the dictionary building above

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Nice, I didn't know logically concatenated statements evaluated iteratively like that! Usually, it's the call to `metadata['Ecephys']` itself that triggers the issue of a missing key.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -758,20 +789,99 @@ def write_sorting(sorting, save_path, **nwbfile_kwargs):
...
-                set_dynamic_table_property(
-                    dynamic_table=nwbfile.units,
-                    row_ids=ids,
-                    property_name=ft,
-                    values=flatten_vals,
-                    index=spikes_index,
-                )
+                values = []
+                skip_feature = False
+                if not ft.endswith('_idxs'):
```
alejoe91, line 796: Refractored the features part a bit. A commn case is to compute features only on a subset of spikes. In that case, SI creates a feature called e.g. `waveforms` and one called `waveforms_idxs`, containing the indexes. Of course the latter case was causing errors, so for let's skip those features in favor of *full* features.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Good call!

spikeextractors/extractors/nwbextractors/nwbextractors.py
bendichter: yes, Python "or" statements have "short-circuit evaluation," so if the left side evaluates as True, the right side is never evaluated. Pretty handy in cases like this!

spikeextractors/extractors/nwbextractors/nwbextractors.py
bendichter: yeah lets proceed

bendichter:
this looks good to me

PR #448, opened 2020-09-08
Neuroscope clu suffix bug fix

CodyCBakerPhD:
Discovered some new special cases for noisy .res and .clu files in a common directory with the files we actually want to extract from. Also greater enforcement of PEP8 and Numpy docstrings.

### DIFF
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -193,9 +193,9 @@ class NeuroscopeSortingExtractor(SortingExtractor):
             assert folder_path.is_dir(), 'The folder_path must be a directory!'
 
             res_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.res' in f.name and '.temp.' not in f.name]
+                         and '.res' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
             clu_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.clu' in f.name and '.temp.' not in f.name]
+                         and '.clu' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
 
             assert len(res_files) > 0 or len(clu_files) > 0, \
                 'No .res or .clu files found in the folder_path.'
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -193,9 +196,9 @@ def __init__(self, resfile_path: PathType = None, clufile_path: PathType = None,
             assert folder_path.is_dir(), 'The folder_path must be a directory!'
 
             res_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.res' in f.name and '.temp.' not in f.name]
+                         and '.res' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
```
CodyCBakerPhD, line 199: The newly discovered format is of the type 'session_id.res.%d.MM.DD.YYYY.HH.MM', but also exists alongside the actual well-formatted 'session_id.res.%d' that we want.

The new error check covers the previous '.temp' case as well as this.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -232,13 +235,10 @@ def __init__(self, resfile_path: PathType = None, clufile_path: PathType = None,
             n_clu = clu[0]
             clu = np.delete(clu, 0)
             unique_ids = np.unique(clu)
-
-            if not np.sort(unique_ids) == np.arange(n_clu + 1):  # some missing IDs somewhere
-                if 0 not in unique_ids:  # missing unsorted IDs
-                    n_clu += 1
-                if 1 not in unique_ids:  # missing mua IDs
-                    n_clu += 1
-                # If it is any other ID, then it would be very strange if it were missing...
```
CodyCBakerPhD, line 241: After a closer look at functional testing on the actual data, the outer `if` statement triggers 100% of the time, and also generally isn't needed in addition to the inner `if` statements.

alejoe91:
Hi @CodyCBakerPhD 

This looks good to me :) great that you've spotted it! 
Let me know if it's ready to merge or if you want to run some more tests.

CodyCBakerPhD:
Unfortunately, the only way to fully test it is to run it over the entire dataset, which we are still building up to. Would it be too annoying to keep putting up a new PR every time we catch a new case? If not this is good to go.

Thankfully the new format detection is much more robust in terms of adherence to the '.res.%d' format, so it should be OK. But nothing is ever truly perfect, either...

PR #449, opened 2020-09-08
Implement get_ttl_events method for available extractors

alejoe91:
This PR implements the `get_ttl_events` method, which returns the frame time of digital channels used for synchronization.

The method signature is : `get_ttl_events(self, start_frame=None, end_frame=None, channel_id=0)`, where `channel_id` is the channel of the digital signal to be returned. 

The method returns:
- ttl_frames: np.array with frames at which TTL occur
- ttl_state: np.array of the same length with +1 (rising) and -1 (falling) 

Currently implemented in:
- SubRecordingExtractor
- MultiRecordingTimeExtractor
- NumpyRecrdingExtractor (for adde tests)
- SpikeGLXRecordingExtractor
- OpenEphysRecordingExtractor
- IntanRecordingExtractor
- Mea1kRecordingExtractor
- MaxOneRecordingExtractor

### DIFF
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -785,22 +785,22 @@ def check_get_ttl_args(func):
             recording = args[0]
             start_frame = kwargs.get('start_frame', None)
             end_frame = kwargs.get('end_frame', None)
-            channel = kwargs.get('channel', 0)
+            channel_id = kwargs.get('channel_id', 0)
         elif len(args) == 2:
             recording = args[0]
             start_frame = args[1]
             end_frame = kwargs.get('end_frame', None)
-            channel = kwargs.get('channel', 0)
+            channel_id = kwargs.get('channel_id', 0)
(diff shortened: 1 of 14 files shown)


samuelgarcia:
Hi Alessio.
Can we have a call for this API extention ?

colehurwitz:
Should we make: channel --> channel_id/channel_ids to be consistent?

colehurwitz:
I have been quite busy with other work, but will try to be more helpful on github as well!

alejoe91:
@samuelgarcia Neo extractors should also implement this! But we can add it later. For me this is ready to be merged

samuelgarcia:
@alejoe91 : I am not super happy ith the happy. I hink we should have 2 layer to get theses ttl frames one would be to have the boolean time series and the other would be the have the up and down edge (like you did here). Give by defult the edge can have super big cost for long files.
Could we have a discussion at least for the API ?

PR #459, opened 2020-10-06
Heavy refactor of NwbRecordingExtractor write_recording()

CodyCBakerPhD:
Starting with a refactor of warnings to assertions in add_devices, @bendichter is this first commit in line with what you were thinking? If so I'll take the same approach to the other auxiliary functions.

@alejoe91 any thoughts?

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -313,7 +313,7 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                     Device=[defaults]
                 )
             )
-
+        print("\n\n" + metadata + "\n\n")
         assert all([isinstance(x, dict) for x in metadata['Ecephys']['Device']]), \
             "Expected metadata['Ecephys']['Device'] to be a list of dictionaries!"
 
@@ -349,46 +349,43 @@ class NwbRecordingExtractor(se.RecordingExtractor):
             assert isinstance(nwbfile, NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
         defaults = {'name': 'Electrode Group',
                     'description': 'no description',
-                    'location': 'unknown'}
-        default_dev_name = 'Device'
-
+                    'location': 'unknown',
+                    'device_name': 'Device'}
         if metadata is None:
-            metadata = dict()
+            metadata = dict(
+                Ecephys=dict(
+                    ElectrodeGroup=[defaults]
+                )
+            )
 
-        if len(metadata.keys()) > 0:
-            if 'Ecephys' in metadata and 'ElectrodeGroup' in metadata['Ecephys']:
-                if type(metadata['Ecephys']['ElectrodeGroup']) is list and metadata['Ecephys']['ElectrodeGroup']:
-                    for j, grp in enumerate(metadata['Ecephys']['ElectrodeGroup']):
-                        if type(grp) is dict:
-                            if grp.get('name', defaults['name']) not in nwbfile.electrode_groups:
-                                if grp.get('device_name', default_dev_name) not in nwbfile.devices:
-                                    new_device = {'Ecephys': {'Device': {'name':
-                                                                         grp.get('device_name', default_dev_name)}}}
-                                    se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
-                                    print("Warning: device name not detected in attempted link to electrode group! "
-                                          "Automatically generating.")
+        assert all([isinstance(x, dict) for x in metadata['Ecephys']['ElectrodeGroup']]), \
+            "Expected metadata['Ecephys']['ElectrodeGroup'] to be a list of dictionaries!"
 
-                                electrode_group_kwargs = dict(defaults, **grp)
-                                electrode_group_kwargs.pop('device_name')
-                                electrode_group_kwargs.update(
-                                    {'device': nwbfile.devices[grp.get('device_name', default_dev_name)]})
-                                nwbfile.create_electrode_group(**electrode_group_kwargs)
-                        else:
-                            print(f"Warning: Expected metadata['Ecephys']['ElectrodeGroup'][{j}] to be"
-                                  " a dictionary with keys 'name', 'description', 'location', and 'device'!"
-                                  f"Electrode Group [{j}] will not be created.")
-                else:
-                    print("Warning: metadata must be a list of dictionaries of the form"
-                          " metadata['Ecephys']['ElectrodeGroup'] = [{'name': my_name,"
-                          " 'description': my_description, 'location': electrode_location,"
-                          " 'device_name': my_device_name}, ...]!")
+        for grp in metadata['Ecephys']['ElectrodeGroup']:
+            device_name = grp.get('device_name', defaults['device_name'])
+            if grp.get('name', defaults['name']) not in nwbfile.electrode_groups:
+                if device_name not in nwbfile.devices:
+                    new_device = dict(
+                                    Ecephys=dict(
+                                        Device=dict(
+                                            name=device_name
+                                        )
+                                    )
+                                 )
+                    se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
+                    print("Warning: device name not detected in attempted link to electrode group! "
+                          "Automatically generating.")
+                electrode_group_kwargs = dict(defaults, **grp)
+                electrode_group_kwargs.pop('device_name')
+                electrode_group_kwargs.update(dict(device=nwbfile.devices[device_name]))
+                nwbfile.create_electrode_group(**electrode_group_kwargs)
 
         if not nwbfile.electrode_groups:
             if len(nwbfile.devices.keys()) == 1:
                 device_names = list(nwbfile.devices.keys())
                 device = nwbfile.devices[device_names[0]]
             elif len(nwbfile.devices.keys()) == 0:
-                new_device = {'Ecephys': {'Device': {'name': default_dev_name}}}
+                new_device = {'Ecephys': {'Device': {'name': defaults['device_name']}}}
                 se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
                 device = nwbfile.devices['Device']
             elif len(nwbfile.devices.keys()) > 1:
@@ -407,8 +404,8 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                     nwbfile.create_electrode_group(**electrode_group_kwargs)
             else:
                 nwbfile.add_electrode_groups(**electrode_group_kwargs)
-                print("Warning: No electrode group metadata provided, and no"
-                      " internal group properties set. Generating default device!")
+                print("Warning: No electrode group metadata provided, and no internal group properties set! "
+                      "Generating default device.")
 
     @staticmethod
     def add_electrodes(recording: se.RecordingExtractor, nwbfile=None,
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -125,17 +125,6 @@ def list_get(l, idx, default):
         return default
 
 
-def fill_kwargs_from_defaults(defaults: dict, values: dict = None):
-    kwargs = {}
-    if values is None:
-        for default_property, default_value in defaults.items():
-            kwargs.update({default_property: default_value})
-    else:
-        for default_property, default_value in defaults.items():
-            kwargs.update({default_property: values.get(default_property, default_value)})
-    return kwargs
-
-
 class NwbRecordingExtractor(se.RecordingExtractor):
     extractor_name = 'NwbRecording'
     has_default_locations = True
@@ -294,8 +283,9 @@ class NwbRecordingExtractor(se.RecordingExtractor):
     @staticmethod
     def add_devices(recording: se.RecordingExtractor, nwbfile=None,
                     metadata: dict = None):
-        '''
+        """
         Auxiliary static method for nwbextractor.
+
         Adds device information to nwbfile object.
         Will always ensure nwbfile has at least one device, but multiple
         devices within the metadata list will also be created.
@@ -312,41 +302,35 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                                                   'description': my_description}, ...]
 
         Missing keys in an element of metadata['Ecephys']['Device'] will be auto-populated with defaults.
-        '''
+        """
         if nwbfile is not None:
             assert isinstance(nwbfile, NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
         defaults = {'name': 'Device',
                     'description': 'no description'}
 
         if metadata is None:
-            metadata = {}
-            metadata['Ecephys'] = {}
-            metadata['Ecephys']['Device'] = []
+            metadata = dict()
+            metadata['Ecephys'] = dict()
+            metadata['Ecephys']['Device'] = [dict(defaults)]
 
-        assert len(metadata.keys()) > 0, "Expected metadata must be a non-empty dictionary!"
-        assert 'Ecephys' in metadata, "Expected metadata to have key 'Ecephys'!"
-        assert 'Device' in metadata['Ecephys'], "Expected metadata['Ecephys'] to have key 'Device'!"
-        assert type(metadata['Ecephys']['Device']) is list, "Expected metadata['Ecephys']['Device'] to be a list!"
+        assert all([isinstance(x, dict) for x in metadata['Ecephys']['Device']]), \
+            "Expected metadata['Ecephys']['Device'] to be a list of dictionaries!"
 
-        for j, dev in enumerate(metadata['Ecephys']['Device']):
-            assert type(dev) is dict, f"Expected metadata['Ecephys']['Device'][{j}] to be a dictionary!"
+        for dev in metadata['Ecephys']['Device']:
             if dev.get('name', defaults['name']) not in nwbfile.devices:
                 nwbfile.create_device(**dict(defaults, **dev))
-
-        if not nwbfile.devices:
-            nwbfile.create_device(**defaults)
-            print("Warning: No device metadata provided. Generating default device!")
 
     @staticmethod
     def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
                              metadata: dict = None):
-        '''
+        """
         Auxiliary static method for nwbextractor.
+
         Adds electrode group information to nwbfile object.
         Will always ensure nwbfile has at least one electrode group.
-        Will auto-generate a linked device if the specified name 
+        Will auto-generate a linked device if the specified name
         does not exist in the nwbfile.
-        
+
         Parameters
         ----------
         recording: RecordingExtractor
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -318,28 +320,20 @@ def add_devices(recording: se.RecordingExtractor, nwbfile=None,
...
+        assert len(metadata.keys()) > 0, "Expected metadata must be a non-empty dictionary!"
+        assert 'Ecephys' in metadata, "Expected metadata to have key 'Ecephys'!"
+        assert 'Device' in metadata['Ecephys'], "Expected metadata['Ecephys'] to have key 'Device'!"
+        assert type(metadata['Ecephys']['Device']) is list, "Expected metadata['Ecephys']['Device'] to be a list!"
+
+        for j, dev in enumerate(metadata['Ecephys']['Device']):
+            assert type(dev) is dict, f"Expected metadata['Ecephys']['Device'][{j}] to be a dictionary!"
+            if dev.get('name', defaults['name']) not in nwbfile.devices:
+                device_kwargs = fill_kwargs_from_defaults(defaults, dev)
+                nwbfile.create_device(**device_kwargs)
```
bendichter, line 335: what if metadata was specified as `None`? This will not create any device. Will that be a problem later?

CodyCBakerPhD: In that case, the final lines of

```
 if not nwbfile.devices:
            device_kwargs = fill_kwargs_from_defaults(defaults)
            nwbfile.create_device(**device_kwargs)
```

will trigger, which in the previous format would have always triggered if something went wrong, but in the new format only triggers in the special cases where either metadata was not specified or metadata['Ecephys']['Device'] was empty.

bendichter:
This function is way more verbose than it needs to be:

```python
def fill_kwargs_from_defaults(defaults: dict, values: dict = None):
    kwargs = {}
    if values is None:
        for default_property, default_value in defaults.items():
            kwargs.update({default_property: default_value})
    else:
        for default_property, default_value in defaults.items():
            kwargs.update({default_property: values.get(default_property, default_value)})
    return kwargs
```

That is equivalent to:

```python

def fill_kwargs_from_defaults(defaults: dict, values: dict = None):
    if values is None:
        return defaults
    return dict(defaults, **values)
```

Since this is only used once, I think it is better to not have this as a function

CodyCBakerPhD:
That's a great point on the dictionary instantiation, `dict(default, **values)` is much more clever. `fill_kwargs_from_defaults` was defined as a function since it was used 1-2 times in each auxiliary function, but since it can be done in the one-line` dict()` call I'll start switching all the other calls over to that and remove it entirely.

bendichter:
This function:
https://github.com/SpikeInterface/spikeextractors/blob/11d0d72f282148c1989eef8864119e64914229f9/spikeextractors/extractors/nwbextractors/nwbextractors.py#L82-L90

should be:

```python
def get_nspikes(units_table, unit_id):
    """Returns the number of spikes for chosen unit."""
    check_nwb_install()
    ids = units_table.id[:]
    index = np.where(ids == unit_id)
    assert len(index), ValueError(
            "{} is an invalid unit_id. Valid ids: {}".format(unit_id, ids.tolist()))
    if index == 0:
        return units_table['spike_times_index'].data[index]
    else:
        return units_table['spike_times_index'].data[index] - units_table['spike_times_index'].data[index - 1]
```
This avoids reading the entire `units_table['spike_times_index']` dataset

bendichter:
Dictionaries should be declared as literals where possible. Instead of

https://github.com/SpikeInterface/spikeextractors/blob/11d0d72f282148c1989eef8864119e64914229f9/spikeextractors/extractors/nwbextractors/nwbextractors.py#L235-L242

This would be better:

```python
self.nwb_metadata = dict(
    NWBFile=dict(
        session_description=nwbfile.session_description,
        identifier=nwbfile.identifier,
        session_start_time=nwbfile.session_start_time,
        institution=nwbfile.institution,
        lab=nwbfile.lab
    )
)
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -310,39 +310,31 @@ def add_devices(recording: se.RecordingExtractor, nwbfile=None,
...
-                        print(f"Warning: Expected metadata['Ecephys']['Device'][{j}] to be"
-                              " a dictionary with keys 'name' and 'description'!"
-                              f" Device [{j}] will not be created.")
-            else:
-                print("Warning: metadata must be a list of dictionaries of the form"
-                      " metadata['Ecephys']['Device'] = [{'name': my_name,"
-                      " 'description': my_description}, ...]!")
+        assert len(metadata.keys()) > 0, "Expected metadata must be a non-empty dictionary!"
+        assert 'Ecephys' in metadata, "Expected metadata to have key 'Ecephys'!"
+        assert 'Device' in metadata['Ecephys'], "Expected metadata['Ecephys'] to have key 'Device'!"
```
bendichter, line 328:
```suggestion
```
Just the last check should suffice

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -310,39 +310,31 @@ def add_devices(recording: se.RecordingExtractor, nwbfile=None,
...
+        assert type(metadata['Ecephys']['Device']) is list, "Expected metadata['Ecephys']['Device'] to be a list!"
+
+        for j, dev in enumerate(metadata['Ecephys']['Device']):
+            assert type(dev) is dict, f"Expected metadata['Ecephys']['Device'][{j}] to be a dictionary!"
+            if dev.get('name', defaults['name']) not in nwbfile.devices:
+                nwbfile.create_device(**dict(defaults, **dev))
```
bendichter, line 334:
```suggestion
        assert isinstance(x, dict) for x in metadata['Ecephys']['Device'], "Expected metadata['Ecephys']['Device'] to be a list of dicts"

        for dev in metadata['Ecephys']['Device']:
            if dev.get('name', defaults['name']) not in nwbfile.devices:
                nwbfile.create_device(**dict(defaults, **dev))
```

bendichter: This single assertion will do the work of all of them

bendichter: `isinstance` should be used instead of `type(x) == Type`

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -310,51 +300,37 @@ def add_devices(recording: se.RecordingExtractor, nwbfile=None,
...
+        """
         if nwbfile is not None:
             assert isinstance(nwbfile, NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
         defaults = {'name': 'Device',
                     'description': 'no description'}
 
         if metadata is None:
             metadata = dict()
+            metadata['Ecephys'] = dict()
+            metadata['Ecephys']['Device'] = [dict(defaults)]
```
bendichter, line 314:
```suggestion
            metadata = dict(
                Ecephys=dict(
                    Device=[defaults]
                )
            )
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -517,12 +492,11 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None,
                           " the nwbfile electrode ids and RecordingExtractor channel ids!")
 
         for j, channel_id in enumerate(channel_ids):
-            # Will not overwrite the electrode id if already in nwbfile
             if channel_id not in nwb_elec_ids:
-                electrode_kwargs = fill_kwargs_from_defaults(defaults)
+                electrode_kwargs = dict(defaults)
```
bendichter, line 496: why use `dict` here?

CodyCBakerPhD: To make a copy (or a new dictionary object) filled with default values, in case operations (like `.pop()` or `.update()`) need to be performed on it for some reason afterward.

bendichter: ah, it's used as a copy. Makes sense, thanks for the explanation :-)

CodyCBakerPhD: Np, it's really only necessary at like one or two parts of the code that actually arise as byproducts of the removal of the previous `fill_kwargs_from_defaults` function, which is actually subtly different from `dict(defaults, **values)` in that it only looped over keys in defaults, so extra keys in values would not actually end up in the final `kwargs`, so in the new method it just has to be popped off before going into the `add_*` function (see line 373 for one of the only occurrences; it may even disappear as I refactor the other functions to match the new style).

Thanks again for all the suggestions! I'll try to get this done by tomorrow-ish pending the progress of the upload speed on that other DANDI set.

alejoe91:
Thanks @CodyCBakerPhD the code is much nicer now. Tests are failing though. I'll look into it :)

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -349,46 +349,43 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
...
+                    se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
+                    print("Warning: device name not detected in attempted link to electrode group! "
+                          "Automatically generating.")
+                electrode_group_kwargs = dict(defaults, **grp)
+                electrode_group_kwargs.pop('device_name')
+                electrode_group_kwargs.update(dict(device=nwbfile.devices[device_name]))
+                nwbfile.create_electrode_group(**electrode_group_kwargs)
 
         if not nwbfile.electrode_groups:
             if len(nwbfile.devices.keys()) == 1:
```
bendichter, line 384:
```suggestion
            if len(nwbfile.devices) == 1:
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -349,46 +349,43 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
...
                 device_names = list(nwbfile.devices.keys())
                 device = nwbfile.devices[device_names[0]]
             elif len(nwbfile.devices.keys()) == 0:
```
bendichter, line 387:
```suggestion
            elif len(nwbfile.devices) == 0:
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -349,46 +349,43 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
...
-                new_device = {'Ecephys': {'Device': {'name': default_dev_name}}}
+                new_device = {'Ecephys': {'Device': {'name': defaults['device_name']}}}
                 se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
                 device = nwbfile.devices['Device']
             elif len(nwbfile.devices.keys()) > 1:
```
bendichter, line 391:
```suggestion
            elif len(nwbfile.devices) > 1:
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: All parts of this are now outdated, code block will always create a device with defaults if the appropriate keys are left blank or  metadata is not passed.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Resolved; still replacing all {} with appropriate dict() below line 380

CodyCBakerPhD:
@bendichter @alejoe91 Ready for another round of reviews.

All instances of dictionary instantiation have been refactored to literals except for line 521

```
if metadata_column['name'] in defaults:
                            electrode_kwargs.update({
                                metadata_column['name']: list_get(metadata_column['data'], j,defaults[metadata_column['name']])
                            })
```

since dict(...) wouldn't allow the key name to be a variable.

All `add_*` functions now follow the format

```
> finish argument type checking (mostly for `NWBFiles`)
> check if necessary upstream link objects have been created (for example, electrode_groups must be linked to devices, electrodes linked to electrode_groups, etc...)
> populate defaults for that object type
> auto-populate correct metadata structure, including defaults in several cases
> assertion checks on the structure of the metadata, removing the need for multiple `if` checks later in the loops
> do the operation to add the object to the `NWBFile`
```

Some things yet remain to be discussed, such as the use cases of `add_electrodes `which involve pathological cases of passed metadata which ultimately result in that electrode not being added to the file; these are the two cases that end in `continue`.

Also, I was wondering if a helper function might be useful for generating default values for each type, something like

```
def generate_defaults(object_type):
   if object_type == "device":
      defaults = dict(...)
   elif object_type == ...
   ...
   return(defaults)
```

might be nice to centralize the use. This is especially so because lines 504-512 partially utilize defaults upstream of the object type at that level, and so rely on their values being the same (as least to keep things consistent).

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Ah, right you are. The `len(dict(....))` returns the # of key-value pairs, so the keys themselves are not necessary for those types of checks. Fixed now

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: All occurrences of `type `ought to now be replaced by `isinstance`

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -367,81 +340,76 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
...
+                    new_device = dict(
+                        Ecephys=dict(
+                            Device=dict(
+                                name=device_name
+                            )
+                        )
+                     )
+                    se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
+                    print("Warning: device name not detected in attempted link to electrode group! "
+                          "Automatically generating.")
```
bendichter, line 381:
```suggestion
                    warnings.warn"Warning: device name {} not detected in attempted link to electrode group! "
                          "Automatically generating.".format(device_name))
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -367,81 +340,76 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
...
+                nwbfile.create_electrode_group(**electrode_group_kwargs)
+
+        if 'group' in recording.get_shared_channel_property_names():
+            RX_groups_names = np.unique(recording.get_channel_groups()).tolist()
+
+            if len(nwbfile.devices) == 0:
+                se.NwbRecordingExtractor.add_devices(recording, nwbfile)
+            elif len(nwbfile.devices) > 1:
+                print("Warning: more than one device found when adding electrode group via channel properties: "
+                      "using first added device. To use a different device, indicate it the metadata argument.")
```
bendichter, line 394: would be nice if this warning message
1. Used `warnings.warn`
2. Indicated the name of the device that was being used. The "first" key may or may not be deterministic, depending on the version of python being used.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -367,81 +340,76 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
...
+            device_names = list(nwbfile.devices.keys())
+            device = nwbfile.devices[device_names[0]]
+
+            electrode_group_kwargs = dict(defaults)
+            electrode_group_kwargs.pop('device_name')
+            electrode_group_kwargs.update(dict(device=device))
```
bendichter, line 400:
```suggestion
            electrode_group_kwargs.update(device=device)
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -367,81 +340,76 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None,
...
+
+            for grp_name in RX_groups_names:
+                electrode_group_kwargs.update(dict(name=str(grp_name)))
```
bendichter, line 403:
```suggestion
                electrode_group_kwargs.update(name=str(grp_name))
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -455,141 +423,148 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None,
...
-        for j, channel_id in enumerate(channel_ids):
-            # Will not overwrite the electrode id if already in nwbfile
+            metadata = dict(
+                Ecephys=dict(
+                    Electrodes=[]
+                )
+            )
+        assert all([isinstance(x, dict) and set(x.keys()) == set(['name', 'description', 'data'])
+                    and isinstance(x['data'], list) for x in metadata['Ecephys']['Electrodes']]), \
+            "Expected metadata['Ecephys']['Electrodes'] to be a list of dictionaries!"
```
bendichter, line 470: If input validation is getting that complex, it might be better to just use a json schema validator

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -455,141 +423,148 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None,
...
+
+        if nwbfile.electrodes is None:
+            nwb_elec_ids = []
+        else:
+            nwb_elec_ids = nwbfile.electrodes.id.data[:]
+        for metadata_column in metadata['Ecephys']['Electrodes']:
+            if metadata_column['name'] not in nwbfile.electrodes.colnames:
+                nwbfile.add_electrode_column(str(metadata_column['name']),
+                                             str(metadata_column['description']))
```
bendichter, line 479: Why wrap these in strings? Are there times when these are not strings?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -455,141 +423,148 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None,
...
+
+        for j, channel_id in enumerate(recording.get_channel_ids()):
             if channel_id not in nwb_elec_ids:
-                electrode_kwargs = fill_kwargs_from_defaults(defaults)
-                electrode_kwargs.update({'id': channel_id})
+                electrode_kwargs = dict(defaults)
+                electrode_kwargs.update(dict(id=channel_id))
```
bendichter, line 484:
```suggestion
                electrode_kwargs.update(id=channel_id)
```

bendichter:
Thanks @CodyCBakerPhD! This code is looking much better. In the future, it would be better to separate this into several PRs.

I think there's a better way to provide defaults than putting them in one big function. Really, the code should expect the metadata to follow the structure of the metadata json schema. We should put validation at the start to ensure that it does. Dealing with non-compliant metadata schemas is making this class really complicated and populating missing values is a separate concern.

CodyCBakerPhD:
Since there is more refactoring needed to meet the remaining suggestions, I do think I'll just break it up into several PRs to make it easier to process.

@bendichter So you're saying I can go ahead and make the assumption that all metadata is properly formatted when passed into this class?

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Gotcha, no longer referring to it as "first" device then, just by whatever its name is (see the other PR for changes).

alejoe91:
Looks good @CodyCBakerPhD 

Merging!

PR #463, opened 2020-10-09
Refactor of add_electrodes in NwbRecordingExtractor

CodyCBakerPhD:
1. Replaced all nested if checks with single line assertion
2. Better organized initial metadata auto-population and defaults, including auto-creation of default device
3. Replaced dictionary definitions with literals
4. Replaced print("Warning: ...") with warnings.warn(...)
5. Enforced Numpy docstring/PEP8

This is by far the biggest refactor of all these `add_*` functions, given the complexity involved due to electrodes being the most downstream component of the general device metadata.

Some of these checks to metadata format may be overkill pending the decision on what to assume about passed structure, so may yet be further changed.


CodyCBakerPhD:
@bendichter Any final thoughts on the electrodes part of the refactor?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -455,141 +456,148 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None,
...
-                    if type(metadata_column) is dict \
-                            and set(metadata_column.keys()) == set(['name', 'description', 'data']) \
-                            and type(metadata_column['data']) is list:
-                        if metadata_column['name'] == 'group':
-                            if list_get(metadata_column['data'], j, np.nan) not in nwbfile.electrode_groups:
-                                print(
-                                    f"Warning: Electrode group for electrode {channel_id} was not found in the nwbfile."
-                                    " Automatically adding!")
-                                se.NwbRecordingExtractor.add_electrode_groups(recording, nwbfile, metadata)
+                if any([loc is np.nan for loc in location]):
```
alejoe91, line 521: @CodyCBakerPhD Here we want both elements of `location` not to be `nan` to update the dictionary.

Also `loc is np.nan` is not the correct way to test for nans. Try this snippet:
```
location = np.array([np.nan, np.nan])
print(location[0] is np.nan)  # --> prints False...
print(np.isnan(location[0])  # --> correctly prints True
```

```suggestion
                if all([not np.isnan(loc) for loc in location]):
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -455,141 +456,148 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None,
...
                 descr = 'no description'
-                # property 'brain_area' of RX channels corresponds to 'location' of NWB electrodes
                 if channel_prop_name == 'brain_area':
-                    if 'location' in nwbfile.electrodes.colnames:
-                        continue
-                    else:
-                        channel_prop_name = 'location'
-                        descr = 'brain area location'
+                    channel_prop_name = 'location'
+                    descr = 'brain area location'
```
alejoe91, line 600: So currently this does not allow to save properties that are not shared among all channels. Should we keep it that way? Otherwise, we could add `np.nans` for columns that are missing that property

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Interestingly, if you remove the `np.array` converion on the list of nans, it works as expected; but I did not realize that `get_channel_locations()` returns things as `np.array`'s, so definitely implementation that then.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: If we decide not to keep things that way, it would probably require a more holistic refactor of the way we treat properties in the Recording and Sorting Extractors. For now let's table that. Upon usage of `write_recording` and `write_sorting`, I currently assume such data has been completed by the user (i.e., if there is missing data for any channel or unit, it has been set to some default value by the user upon establishing the properties for that extractor object).

The current refactor shouldn't change the way it was previously handled, which again assumes complete shared data for the internal properties.

bendichter:
looks good to me!

PR #465, opened 2020-10-13
MultiRecordingTimeExtractor update

alejoe91:
Currently the `MultiRecordingTimeExtractor` only allows set of recordings with the same channel ids.

A nice use of this is if there are multiple electrode configurations with some shared channels. In that case, the channel ids of the shared channels would be different, but the locations MUST be the same. 

This PR implements that (and retrieves the correct traces accordingly)

### DIFF
--- a/spikeextractors/multirecordingtimeextractor.py
+++ b/spikeextractors/multirecordingtimeextractor.py
@@ -12,10 +12,12 @@ class MultiRecordingTimeExtractor(RecordingExtractor):
         self._first_recording = recordings[0]
         self._num_channels = self._first_recording.get_num_channels()
         self._channel_ids = self._first_recording.get_channel_ids()
+        self._channel_ids_dict = {}
+
         self._sampling_frequency = self._first_recording.get_sampling_frequency()
         if epoch_names is None:
             epoch_names = [str(i) for i in range(len(recordings))]
-            
+
         RecordingExtractor.__init__(self)
 
         # Add all epochs to the epochs data structure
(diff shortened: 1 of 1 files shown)


alejoe91:
@colehurwitz can you check this one?

spikeextractors/multirecordingtimeextractor.py
```diff
@@ -26,14 +28,26 @@ def __init__(self, recordings, epoch_names=None):
...
-                raise ValueError("Inconsistent channel ids between extractor 0 and extractor " + str(i + 1))
-            if (self._sampling_frequency != sampling_frequency):
-                raise ValueError("Inconsistent sampling frequency between extractor 0 and extractor " + str(i + 1))
+            assert len(self._channel_ids) == len(channel_ids), f"Inconsistent number of channels between " \
+                                                               f"extractor 0 and extractor {i + 1}"
+            assert self._sampling_frequency == sampling_frequency, f"Inconsistent sampling frequency between " \
+                                                                   f"extractor 0 and extractor {i + 1}"
+            assert np.array_equal(self._first_recording.get_channel_locations(), recording.get_channel_locations()), \
+                f"Inconsistent locations between extractor 0 and extractor {i + 1}"
+            if self._channel_ids != channel_ids:
```
samuelgarcia, line 42: This looks like very very permissive. What is the suer case ?

alejoe91: We use several channel configuration of 1000 channels but only 100-200 shared channels covering hot spots. I need to sort those 200 channels and concatenate them

samuelgarcia: You should concatenate 2 subrecording with the aligned channel_ids so. No ?
Making hidden magic is not a good practice, I think.

alejoe91: No. They have different chanel ids because they have different configurations, but shared channels with the same location

samuelgarcia: I have an internal warning flag that is flashing in red and telling me that must be done in two steps with 2 distinct functions(classes) : 1. align/select channel with same localtion 2. concatenate in time.
I know your tenacity : you will convice me but at the moment it is not the case.

alejoe91: of course I do that externally! I think the use case is so niche that it shouldn't make it into the class...

alejoe91:
Using `SubRecordingExtractor` and resetting channel ids (`renamed_channel_ids`) seems to work instead

samuelgarcia:
Very wise.

PR #468, opened 2020-10-15
Allows SpikeGLXRecordingExtractor to extract LFP files

emilyasterjones:
Adds an optional flag (lfp) to the parameters which enables it to parse lf.bin files properly. Without this modification, the current __init__ function only looks for AP channels, and will assume there are 0 channels of data in an lf.bin file. Tested on my own ap.bin and lf.bin files and it works.

### DIFF
--- a/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py
+++ b/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py
@@ -13,7 +13,7 @@ class SpikeGLXRecordingExtractor(RecordingExtractor):
     mode = 'file'
     installation_mesg = ""  # error message when not installed
 
-    def __init__(self, file_path, x_pitch=21, y_pitch=20):
+    def __init__(self, file_path, lfp=False, x_pitch=21, y_pitch=20):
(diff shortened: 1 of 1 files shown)


alejoe91:
Hi @emilyasterjones 

Thank for the PR. The `npx_file` can already be either the ap or the lp one (https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py#L22).

Instead of using an extra argument, you could the `lfp` flag as follows:
```
if 'ap' in self._npxfile.name:
    lfp=False
    self.is_filtered = True
else:
    assert 'lf' in self._npxfile.name, "The 'npx_file should be either the 'ap' or the 'lf' file."
    lfp = True
```

Could you also add this assertion?
`assert 'bin' in self._npxfile.name, "The 'npx_file should be either the 'ap' or the 'lf' bin file."`

The rest looks good :)

emilyasterjones:
Done! Thanks for the coding suggestion assist. Definitely a more elegant solution than requiring a parameter be specified.

alejoe91:
Thanks! merging :)

PR #470, opened 2020-10-16
refactor to helper function get_nspikes used in write_units

CodyCBakerPhD:
Suggested by @bendichter on the [previous PR](https://github.com/SpikeInterface/spikeextractors/pull/459). 

Only change required to pass tests is the conversion of `units_table.id[:]` from `list` to `np.array` in order for the `ids == unit_id` to evaluate as intended.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -85,7 +85,8 @@ def get_nspikes(units_table, unit_id):
     check_nwb_install()
     ids = np.array(units_table.id[:])
     indexes = np.where(ids == unit_id)[0]
-    assert len(indexes), ValueError(f"{unit_id} is an invalid unit_id. Valid ids: {ids}.")
+    if not len(indexes):
+        raise ValueError(f"{unit_id} is an invalid unit_id. Valid ids: {ids}.")
     index = indexes[0]
     if index == 0:
         return units_table['spike_times_index'].data[index]
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -81,14 +81,16 @@ def get_dynamic_table_property(dynamic_table, *, row_ids=None, property_name):
 
 
 def get_nspikes(units_table, unit_id):
-    """Returns the number of spikes for chosen unit."""
+    """Return the number of spikes for chosen unit."""
     check_nwb_install()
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -81,14 +81,16 @@ def get_dynamic_table_property(dynamic_table, *, row_ids=None, property_name):
...
     check_nwb_install()
-    if unit_id not in units_table.id[:]:
-        raise ValueError(str(unit_id) + " is an invalid unit_id. "
-                                        "Valid ids: " + str(units_table.id[:].tolist()))
-    nSpikes = np.diff([0] + list(units_table['spike_times_index'].data[:])).tolist()
-    ind = np.where(np.array(units_table.id[:]) == unit_id)[0][0]
-    return nSpikes[ind]
+    ids = np.array(units_table.id[:])
+    indexes = np.where(ids == unit_id)[0]
+    assert len(indexes), ValueError(f"{unit_id} is an invalid unit_id. Valid ids: {ids}.")
```
CodyCBakerPhD, line 88:
```suggestion
    assert len(indexes), ValueError(f"{unit_id} is an invalid unit_id. Valid ids: {ids}.")
```

Perhaps it is specific to Windows, or my IDE, but whenever I write assertions of the form `assert condition, TypeError(message)`, the final returned error is the `AssertionError`, not the desired `TypeError`. This would be relevant if anything tried to rely on this this function in a `try: except ValueError:` manner. To actually have it return the `ValueError`, this would have to be

```
if not len(indexes):
    raise ValueError(f"{unit_id} is an invalid unit_id. Valid ids: {ids}.")
```

bendichter: @CodyCBakerPhD you're right, the correct syntax is:

```python
assert 1==2, 'error message'
```

if you need another error type, it's better to keep it as:

```python
if error_condition:
    raise ValueError('error message')
```

CodyCBakerPhD:
@alejoe91 While the tests are passing, I do want to double check it runs fine on actual files. I'll let you know if it's good in a little bit.

CodyCBakerPhD:
@alejoe91 OK, output looks good for merge. This change should result in a decent decrease to the computational complexity by not having to read the entire units data table.

alejoe91:
> @alejoe91 OK, output looks good for merge. This change should result in a decent decrease to the computational complexity by not having to read the entire units data table.

Thanks Cody! I agree it looks much better :) merging

PR #473, opened 2020-10-19
simple fix to add_electrode_group functionality

CodyCBakerPhD:
I should have taken a closer look at this before OK'ing the last one for merge; after taking another look at the logic, as well as the default behavior of functions like `recording.get_shared_channel_property_names()`, I've removed some unnecessary if checks that were always tautologically getting called or ignored.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -360,6 +360,8 @@ class NwbRecordingExtractor(se.RecordingExtractor):
         """
         if nwbfile is not None:
             assert isinstance(nwbfile, NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
+        if len(nwbfile.devices) == 0:
+            se.NwbRecordingExtractor.add_devices(recording, nwbfile)
         defaults = dict(
             name="Electrode Group",
             description="no description",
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -396,9 +396,7 @@ def add_electrode_groups(recording: se.RecordingExtractor, nwbfile=None, metadat
                 electrode_group_kwargs.update(device=nwbfile.devices[device_name])
                 nwbfile.create_electrode_group(**electrode_group_kwargs)
 
-        if 'group' in recording.get_shared_channel_property_names():
-            if len(nwbfile.devices) == 0:
-                se.NwbRecordingExtractor.add_devices(recording, nwbfile)
+        if not nwbfile.electrode_groups:
```
CodyCBakerPhD, line 399: This one line addition also returns the behavior back to where it was prior to the refactor merged a couple days ago; that is, we only rely upon the internal 'group' channel property if no metadata results in the creation of devices in the previous code block.

This is largely because there is, in fact, always a 'group' property, and without using the `metadata` to somehow set this  (which since it can only be an integer in the `RecordingExtractor`, loses some metadata information about the desired name as a string), it will default to all channels belonging to group `0`

PR #481, opened 2020-11-02
Make BinDat get_traces() return memmap; add compression to NWB write_recording()

CodyCBakerPhD:
- Significant changes to add_electrical_series in nwbRecordingExtractor
- Fix to memmap return types by bindatRecordingExtractor
- Some style enforcement (PEP8/Numpy, 120 space cutoff)
- Small hotfix to `add_electrodes()`
- Added compression to all DataChunkIterators in NWB

@alejoe91 @bendichter Very significant (though thankfully small) changes to the `add_electrical_series()` pipeline going all the way down to the BinDatRecording level. These changes have successfully allowed processing on binary files in excess of 250GB, where the channel ids were subset using a SubRecordingExtractor. Previously, the data generator part of `add_electrical_series()` returned an entire array of the very long recording, causing a remote server with a high RAM cap to force termination of the process.

Due to it's importance to a number of active projects, I'd recommend this going through ASAP after your approval, but for the long term there are two additional avenues to further improving this functionality over (a) non-binary files, and (b) binary files where the subsetting is not consecutive as we assume in the case of this PR.

(a) can be solved either by extending the current DataChunkIterator over on hdmf, or by making our own that is intended to read/append matrix valued chunks of matrix valued data.

(b) can be potentially be solved by something like [this stack overflow discussion](https://stackoverflow.com/questions/13780907/is-it-possible-to-np-concatenate-memory-mapped-files).


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -485,7 +486,7 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None, metadata: dic
         else:
             nwb_elec_ids = nwbfile.electrodes.id.data[:]
         for metadata_column in metadata['Ecephys']['Electrodes']:
-            if metadata_column['name'] not in nwbfile.electrodes.colnames:
+            if nwbfile.electrodes is None or metadata_column['name'] not in nwbfile.electrodes.colnames:
```
CodyCBakerPhD, line 489: This is the hotfix to add_electrodes() when running on a fresh or defaulted NWBFile.

spikeextractors/extraction_tools.py
```diff
@@ -257,12 +257,10 @@ def read_binary(file, numchan, dtype, time_axis=0, offset=0):
     with Path(file).open() as f:
         nsamples = (os.fstat(f.fileno()).st_size - offset) // (numchan * np.dtype(dtype).itemsize)
     if time_axis == 0:
-        samples = np.memmap(file, np.dtype(dtype), mode='r', offset=offset,
-                            shape=(nsamples, numchan))
+        samples = np.memmap(file, np.dtype(dtype), mode='r', offset=offset, shape=(nsamples, numchan))
```
CodyCBakerPhD, line 260: Style; maximum line length in SpikeExtractors is 120

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
```diff
@@ -68,7 +68,11 @@ def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
             recordings = self._timeseries[:, start_frame:end_frame]
         else:
             channel_idxs = np.array([self.get_channel_ids().index(ch) for ch in channel_ids])
-            recordings = self._timeseries[:, start_frame:end_frame][channel_idxs, :]
+            if np.all(np.diff(channel_idxs) == 1):
+                recordings = self._timeseries[channel_idxs[0]:channel_idxs[0]+len(channel_idxs), start_frame:end_frame]
+            else:
+                # This block of the execution will return the data as an array, not a memmap
+                recordings = self._timeseries[:, start_frame:end_frame][channel_idxs, :]
```
CodyCBakerPhD, line 75: Indexed memmaps return memmap objects if the subsetting uses only the ":" operator. If a list, np.array, or anything else is passed on that place it will return an array (i.e., access the data and fill it into memory).

A more permanent way of getting rid of that last else statement may be offered by that stack overflow discussion, but I think we should save that for a subsequent PR.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -644,28 +646,43 @@ def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metada
...
-                rate=recording.get_sampling_frequency(),
-                conversion=scalar_conversion,
-                channel_conversion=channel_conversion,
-                comments='Generated from SpikeInterface::NwbRecordingExtractor',
-                description=metadata['Ecephys']['ElectricalSeries'].get('description', defaults['description'])
+            if isinstance(recording.get_traces(), np.memmap):
+                ephys_data = H5DataIO(
+                    DataChunkIterator(
+                        data=np.memmap.transpose(recording.get_traces()),  # nwb standard is time as zero axis
+                        buffer_size=10**6
```
CodyCBakerPhD, line 653: @bendichter @alejoe91 Setting the buffer_size is essentially like setting the chunk size in the data generator below. Too low and the process will take much longer to run than otherwise necessary; too high and it _may_ trigger memory problems.

The 10**6 is currently hardcoded based on my test runs on a remote server; I did some minimal testing locally but don't know how well it scales. Any ideas for better setting this value or determining it from the system calling it?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -644,28 +646,43 @@ def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metada
...
-                starting_time=recording.frame_to_time(0),
-                rate=recording.get_sampling_frequency(),
-                conversion=scalar_conversion,
-                channel_conversion=channel_conversion,
-                comments='Generated from SpikeInterface::NwbRecordingExtractor',
-                description=metadata['Ecephys']['ElectricalSeries'].get('description', defaults['description'])
+            if isinstance(recording.get_traces(), np.memmap):
+                ephys_data = H5DataIO(
+                    DataChunkIterator(
+                        data=np.memmap.transpose(recording.get_traces()),  # nwb standard is time as zero axis
```
alejoe91, line 652: I think that `recording.get_traces().T` will automatically use `np.memmap.transpose` as `recording.get_traces()` is now a memmap object

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: Yes :) 
Check this: https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extraction_tools.py#L269-L316

Here the chunk size is either passed woth the `chunk_size`arg (default `None`) or it's computed based on how much RAM you allow the writing to use (`chunk_mb=500`). I think having something like this could be nice.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: I suspect it would as well, but will triple check. I was basing this syntax off of what's in the read_binary helper function; would you like me to change it there as well?

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Yeah, I remember that function; and actually being able to specific RAM overhead is a really nice feature but haven't looked into how to actually do that. Will do that now.

spikeextractors/extractors/nwbextractors/nwbextractors.py
bendichter: `recording.get_traces().transpose` should work too

CodyCBakerPhD: OK; tested it and both the .transpose and .T worked.

spikeextractors/extraction_tools.py
```diff
@@ -257,8 +257,7 @@ def read_binary(file, numchan, dtype, time_axis=0, offset=0):
     with Path(file).open() as f:
         nsamples = (os.fstat(f.fileno()).st_size - offset) // (numchan * np.dtype(dtype).itemsize)
     if time_axis == 0:
-        samples = np.memmap(file, np.dtype(dtype), mode='r', offset=offset, shape=(nsamples, numchan))
-        samples = np.memmap.transpose(samples)
+        samples = np.memmap(file, np.dtype(dtype), mode='r', offset=offset, shape=(nsamples, numchan)).T
```
CodyCBakerPhD: Changed it here as well

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -587,7 +587,7 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None, metadata: dic
                 )
 
     @staticmethod
-    def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metadata: dict = None):
+    def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metadata: dict = None, buffer_mb=500):
```
CodyCBakerPhD, line 590: I have added the optional buffer_mb as an extra argument here; but reconciling this with the general write_recording(), as well as the nwb-conversion-tools pipeline will take some thought.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -647,10 +647,12 @@ def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metada
                 channel_conversion = gains * 1e-6
 
             if isinstance(recording.get_traces(), np.memmap):
+                n_bytes = np.dtype(recording.get_dtype()).itemsize
+                buffer_size = int(buffer_mb * 1e6) // (recording.get_num_channels() * n_bytes)
                 ephys_data = H5DataIO(
                     DataChunkIterator(
                         data=recording.get_traces().T,  # nwb standard is time as zero axis
-                        buffer_size=10**6
+                        buffer_size=buffer_size
```
CodyCBakerPhD: This ought to mimic that part of the `write_to_binary_dat_format()` function. For 500mb acting on int16, it actually solves for something about 3x larger than the 10**6, so should be a good default.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: @alejoe91 OK, check out that latest commit and comments.

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: Can you call it `chunk_mb` so we keep the same syntax?

CodyCBakerPhD: It's called `buffer` instead of `chunk` in the `DataChunkIterator` that this function actually uses, so I'd prefer to use the convention that the downstream functions use. This function doesn't directly use the `write_to_binary_dat_format()` which is what uses the `chunk` naming convention.

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: Have you added it to the docstring?

CodyCBakerPhD: Nope, forgot; I'll do that now.

CodyCBakerPhD: OK, added to docstring; also added some minimal error checking to it as an (a) 'int' type and (b) at least 10MB. I can imagine a pathological use case where it's set to a low (but non-negative) value such that the floored division in the chunk size calculation returns zero, which would definitely be problematic.

bendichter:
This looks good to me

PR #482, opened 2020-11-05
Speed up copy properties and features when all channels are used

alejoe91:


samuelgarcia:
I think this is not wise to do reference instead copy.
If the original property change so the other is also changed and this is not desired behavior. Even more if this depend to the channel selection.

alejoe91:
I can do copy ;)

samuelgarcia:
cool

PR #484, opened 2020-11-05
Allow extractor module not to have a version (for custom)

alejoe91:
This allows to use parallel processing with custom extractors

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -394,7 +394,6 @@ def _check_same_version(class_string, version):
         return 'unknown'
 
 
-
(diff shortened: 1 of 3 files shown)


alejoe91:
@samuelgarcia the last commit uses `deepcopy` for copying properties and features (related to https://github.com/SpikeInterface/spikeextractors/pull/482#issuecomment-722591709)

samuelgarcia:
Cool.
Is it still fast ?

alejoe91:
> Cool.
> Is it still fast ?

A tiny bit slower than just pointing to the `_properties`, but way faster than looping through all channels and property names.

I'll merge :)

PR #485, opened 2020-11-07
bug fix for intan get_traces()

CodyCBakerPhD:
Basically, when the `NwbRecordingExtractor.write_recording()` reaches [this step](https://github.com/catalystneuro/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L665-L669), the part of `IntanRecordingExtractor.get_traces()` that [indexes channels after it reads them from pyintan](https://github.com/catalystneuro/spikeextractors/blob/master/spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py#L55-L56) throws an indexing issue.

This indexing issue does not occur if your channel indices are consecutive starting from 0, and the current fix as of the submission of this PR has only been confirmed to work when extracting a single index (which is the main use for my current purposes).

### DIFF
--- a/spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py
+++ b/spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py
@@ -52,15 +52,7 @@ class IntanRecordingExtractor(RecordingExtractor):
     def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
         channel_idxs = np.array([self._channel_ids.index(ch) for ch in channel_ids])
         analog_chans = self._analog_channels[channel_idxs]
-        tr = self._recording._read_analog(
-            channels=analog_chans,
-            i_start=start_frame,
-            i_stop=end_frame
-        )
-        if len(channel_idxs) == 1:
-            return tr.T
-        else:
-            return tr[:, channel_idxs].T
+        return self._recording._read_analog(channels=analog_chans, i_start=start_frame, i_stop=end_frame).T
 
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py
```diff
@@ -52,8 +52,15 @@ def get_sampling_frequency(self):
...
-                                            i_start=start_frame, i_stop=end_frame)[:, channel_idxs].T
+        tr = self._recording._read_analog(
+            channels=analog_chans,
+            i_start=start_frame,
+            i_stop=end_frame
+        )
+        if len(channel_idxs) == 1:
+            return tr.T
+        else:
+            return tr[:, channel_idxs].T
```
CodyCBakerPhD, line 63: Actually, why is the secondary indexing even needed? Shouldn't `analog_chans` already have the correct indexing behavior?
```suggestion
        return self._recording._read_analog(channels=analog_chans,
                                            i_start=start_frame, i_stop=end_frame).T
```

alejoe91: Hi Cody! I made some changes recently and might have broken something. Thanks for the fix! I'll test it tomorrow and merge :)

alejoe91:
Hi @CodyCBakerPhD 

Implemented your suggested change and it works well :)

CodyCBakerPhD:
@alejoe91 Glad to hear it! Works well on my end, too.

PR #487, opened 2020-11-08
small changes to NwbRecordingExtractor add_electrical_series

CodyCBakerPhD:
@bendichter If the traces return a non-memory map object, and the data generator is used, the `DataChunkIterator` does not know what the true channel dimension is and so leaves it as `unlimited`. Not a huge issue, but this fixes it.

@alejoe91 Opening as draft because I might include some data conversion type stuff here as well.

Note this branch also includes the `add_electrode` hotfix on the other PR, for testing purposes.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -486,7 +486,8 @@ class NwbRecordingExtractor(se.RecordingExtractor):
         else:
             nwb_elec_ids = nwbfile.electrodes.id.data[:]
         for metadata_column in metadata['Ecephys']['Electrodes']:
-            if nwbfile.electrodes is None or metadata_column['name'] not in nwbfile.electrodes.colnames:
+            if (nwbfile.electrodes is None or metadata_column['name'] not in nwbfile.electrodes.colnames) \
+                    and metadata_column['name'] != 'group_name':
                 nwbfile.add_electrode_column(str(metadata_column['name']),
                                              str(metadata_column['description']))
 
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
@alejoe91 We'll just resolve any future data type conversion changes on a different PR, then. This should be good to go.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -486,7 +486,8 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None, metadata: dic
         else:
             nwb_elec_ids = nwbfile.electrodes.id.data[:]
         for metadata_column in metadata['Ecephys']['Electrodes']:
-            if nwbfile.electrodes is None or metadata_column['name'] not in nwbfile.electrodes.colnames:
+            if (nwbfile.electrodes is None or metadata_column['name'] not in nwbfile.electrodes.colnames) \
+                    and metadata_column['name'] != 'group_name':
```
bendichter, line 490: @CodyCBakerPhD what is the purpose of this?

CodyCBakerPhD: @bendichter See [this branch](https://github.com/SpikeInterface/spikeextractors/pull/486).

CodyCBakerPhD: @bendichter Long term, I'm thinking of more clever ways of checking against restricted column names for the electrodes table prior to it being created; something like the get_method_signature that  we use over on nwb-conversion-tools.

PR #489, opened 2020-11-09
Fix to BinDat retrieve traces when not slices

alejoe91:


samuelgarcia:
Did you benchmark this ? I think it was faster before.
Not sure but need to be tested.

alejoe91:
@samuelgarcia it might be slightly slower, but in case you have a large recording and a subrecording extractor it reads to memory all channels. I found this when dealing with a 20000 electrodes recording and sub-selecting 1000...

samuelgarcia:
I am not talking about the fact you use slice : this is good because under the wood it hack the strides.
I am talking about the fact you slice in the two dimenssion at once.
I think it is better to slice first in time axis and then in channels axis. This was the case explicitly before your change.

PR #490, opened 2020-11-09
Return intan traces in 'float' or 'uint16'

alejoe91:


samuelgarcia:
I think we need something at API level with something more explicit.
`raw_magnitude=True/False`

alejoe91:
> I think we need something at API level with something more explicit.
> `raw_magnitude=True/False`

Another point for SI 2.0

PR #492, opened 2020-11-12
Allowed for overall and single Neuroscope files

wuffi:
Multiple `*.res.*` and `*.clu.*` files (one for each shank) as output of *Klusters* (*Neuroscope*) can potentially be located in the same folder next to a single `*.res` and `*.clu` file containing the data for all shanks.
Moreover, the generation of backup files ending with a tilde like `*.res.*~` by text editors like Emacs can cause failure of the `NeuroscopeSortingExtractor` or `NeuroscopeMultiSortingExtractor` classes.
Finally, `NeuroscopeMultiSortingExtractor` would not accept data, if data was recorded with only one shank, i.e. if there were only one `*.res.1` and `*.clu.1` file (because of assertion in line 406).
The introduced changes make NeuroscopeSortingExtractor only look for *.res and *.clu and NeuroscopeMultiSortingExtractor only look for `*.res.*` and `*.clu.*` while making both extractors ignore files whose name ends with a tilde (`~`).

### DIFF
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -407,8 +407,7 @@ class NeuroscopeMultiSortingExtractor(MultiSortingExtractor):
                      and len(f.suffixes) == 2]
         clu_files = [f for f in folder_path.iterdir() if f.is_file()
                      and '.clu' in f.suffixes 
-                     and '.temp' not in f.suffixes
-                     and not f.name.endswith('~')
+                     re.search(r'\d+$', f.name) is not None
                      and len(f.suffixes) == 2]
 
         assert len(res_files) > 0 or len(clu_files) > 0, \
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -197,9 +197,15 @@ class NeuroscopeSortingExtractor(SortingExtractor):
             assert folder_path.is_dir(), 'The folder_path must be a directory!'
 
             res_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.res' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                         and '.res' in f.suffixes 
+                         and '.temp' not in f.suffixes
+                         and not f.name.endswith('~')
+                         and len(f.suffixes) == 1]
             clu_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.clu' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                         and '.clu' in f.suffixes 
+                         and '.temp' not in f.suffixes
+                         and not f.name.endswith('~')
+                         and len(f.suffixes) == 1]
 
             assert len(res_files) > 0 or len(clu_files) > 0, \
                 'No .res or .clu files found in the folder_path.'
@@ -397,15 +403,21 @@ class NeuroscopeMultiSortingExtractor(MultiSortingExtractor):
             'samplingRate').text)  # careful not to confuse it with the lfpsamplingrate
 
         res_files = [f for f in folder_path.iterdir() if f.is_file()
-                     and '.res' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                     and '.res' in f.suffixes 
+                     and '.temp' not in f.suffixes
+                     and not f.name.endswith('~')
+                     and len(f.suffixes) == 2]
         clu_files = [f for f in folder_path.iterdir() if f.is_file()
-                     and '.clu' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                     and '.clu' in f.suffixes 
+                     and '.temp' not in f.suffixes
+                     and not f.name.endswith('~')
+                     and len(f.suffixes) == 2]
 
         assert len(res_files) > 0 or len(clu_files) > 0, \
             'No .res or .clu files found in the folder_path.'
-        assert len(res_files) > 1 and len(clu_files) > 1, \
-            'Single .res and .clu pairs found in the folder_path. ' \
-            'For single .res and .clu files, use the NeuroscopeSortingExtractor instead.'
+        # assert len(res_files) > 1 and len(clu_files) > 1, \
+        #     'Single .res and .clu pairs found in the folder_path. ' \
+        #     'For single .res and .clu files, use the NeuroscopeSortingExtractor instead.'
         assert len(res_files) == len(clu_files)
 
         res_ids = [int(x.suffix[1:]) for x in res_files]
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from spikeextractors.extraction_tools import check_valid_unit_id, get_sub_extractors_by_property
 from typing import Union
 import os
+import re
 
 try:
     from lxml import etree as et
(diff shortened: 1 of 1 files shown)


alejoe91:
Hi @wuffi, thanks for the PR.

@CodyCBakerPhD can you take a look at this?

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -397,15 +403,21 @@ def __init__(self, folder_path: PathType, keep_mua_units: bool = True, exclude_s
             'samplingRate').text)  # careful not to confuse it with the lfpsamplingrate
 
         res_files = [f for f in folder_path.iterdir() if f.is_file()
-                     and '.res' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                     and '.res' in f.suffixes 
+                     and '.temp' not in f.suffixes
+                     and not f.name.endswith('~')
```
CodyCBakerPhD, line 408: What would you think about adding `re` as an import, and replacing these two lines with

` re.search(r'\d+$', f.name) is not None`

Basically, it just makes these checks a hard enforcement of the .res.%i format. If it ends in ~ or anything else (which has been observed before), it won't pass the check.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -397,15 +403,21 @@ def __init__(self, folder_path: PathType, keep_mua_units: bool = True, exclude_s
...
+                     and len(f.suffixes) == 2]
         clu_files = [f for f in folder_path.iterdir() if f.is_file()
-                     and '.clu' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                     and '.clu' in f.suffixes 
+                     and '.temp' not in f.suffixes
+                     and not f.name.endswith('~')
```
CodyCBakerPhD, line 413: What would you think about adding `re` as an import, and replacing these two lines with

` re.search(r'\d+$', f.name) is not None`

Basically, it just makes these checks a hard enforcement of the .clu.%i format. If it ends in ~ or anything else (which has been observed before), it won't pass the check.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -397,15 +403,21 @@ def __init__(self, folder_path: PathType, keep_mua_units: bool = True, exclude_s
...
+                     and len(f.suffixes) == 2]
 
         assert len(res_files) > 0 or len(clu_files) > 0, \
             'No .res or .clu files found in the folder_path.'
-        assert len(res_files) > 1 and len(clu_files) > 1, \
-            'Single .res and .clu pairs found in the folder_path. ' \
-            'For single .res and .clu files, use the NeuroscopeSortingExtractor instead.'
+        # assert len(res_files) > 1 and len(clu_files) > 1, \
+        #     'Single .res and .clu pairs found in the folder_path. ' \
+        #     'For single .res and .clu files, use the NeuroscopeSortingExtractor instead.'
```
CodyCBakerPhD, line 420: You can go ahead and remove these lines entirely in that case.

CodyCBakerPhD (commented):
See my suggestions. Overall, thanks for bring this to our attention; I had forgotten that other OS add ~ for temporary backups. I had also never personally seen a case where there are non-indexed pairs in the same folder as indexed ones, but I can see how that might occur.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -197,9 +197,15 @@ def __init__(self, resfile_path: PathType = None, clufile_path: PathType = None,
             assert folder_path.is_dir(), 'The folder_path must be a directory!'
 
             res_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.res' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                         and '.res' in f.suffixes 
+                         and '.temp' not in f.suffixes
```
CodyCBakerPhD, line 201:
```suggestion
                         and '.res' in f.suffixes 
```

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -197,9 +197,15 @@ def __init__(self, resfile_path: PathType = None, clufile_path: PathType = None,
...
+                         and not f.name.endswith('~')
+                         and len(f.suffixes) == 1]
             clu_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.clu' in f.suffixes and '.temp' not in f.suffixes and len(f.suffixes) <= 2]
+                         and '.clu' in f.suffixes 
+                         and '.temp' not in f.suffixes
```
CodyCBakerPhD, line 206:
```suggestion
```

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: You can remove the 'temp' option here since the len()==1 negates it

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: You can remove the 'temp' option here since the len()==1 negates it

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD, line 408:
```suggestion
                     re.search(r'\d+$', f.name) is not None
```

CodyCBakerPhD: What would you think about adding re as an import, and replacing these two lines my suggestion?

Basically, it just makes these checks a hard enforcement of the .res.%i format. If it ends in ~ or anything else (which has been observed before), it won't pass the check.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD, line 413:
```suggestion
                     and '.temp' not in f.suffixes
                     and not f.name.endswith('~')
```

CodyCBakerPhD: What would you think about adding re as an import, and replacing these two lines my suggestion?

Basically, it just makes these checks a hard enforcement of the .clu.%i format. If it ends in ~ or anything else (which has been observed before), it won't pass the check.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD, line 420:
```suggestion
```

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
wuffi: Good point. I was hesitant in adding an additional `import`, but the `re` module is standard library.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
wuffi: Strange, I'm of course fine with the change, but can't commit this suggestion because GitHub tells me there would be no changes to the code.

wuffi:
Before the merge is done, I should probably stash the commits into one.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: You can modify it manually on your end of the code (I can't access your branch here) and commit that way

CodyCBakerPhD:
@wuffi Don't forget to add `import re` up top as well as add the missing `and` to those last couple commits.

CodyCBakerPhD:
@alejoe91 Looks good to me, pending the travis-ci approval

PR #495, opened 2020-11-13
Ibl alf sortingextractor

Saksham20:
sorting extractor for reading [ALF ](https://int-brain-lab.github.io/iblenv/one_docs/one_reference.html#alf)format spike sorted files.


spikeextractors/extractorlist.py
```diff
@@ -49,6 +51,7 @@
     SHYBRIDRecordingExtractor,
     NIXIORecordingExtractor,
     NeuroscopeRecordingExtractor,
+    ALFSortingExtractor,
```
alejoe91, line 54: This should go in the sorting extractor list

Saksham20: oh yes, done

PR #498, opened 2020-11-13
Add 'group' check to add_electrode

CodyCBakerPhD:
1) Removed unused auxiliary function.
2) Added assertion to check and instruct users to not pass 'group' as an electrode metadata column any more since it is never used, and entirely overridden in functionality by 'group_name'. The docstring had actually indicated this since the refactor, but it was still possible to pass it in to no effect.


bendichter:
LGTM

PR #502, opened 2020-11-19
Fix copy properties for Sub and Multi extractor

alejoe91:
Fixes https://github.com/SpikeInterface/spiketoolkit/issues/411

### DIFF
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -886,7 +886,7 @@ def cast_start_end_frame(start_frame, end_frame):
         end_frame = end_frame
     else:
         raise ValueError("end_frame must be an int, float (not infinity), or None")
-    return start_frame, end_frame
+    return start_frame.astype('int64'), end_frame.astype('int64')
 
 
 def divide_recording_into_time_chunks(num_frames, chunk_size, padding_size):
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -886,7 +886,11 @@ def cast_start_end_frame(start_frame, end_frame):
         end_frame = end_frame
     else:
         raise ValueError("end_frame must be an int, float (not infinity), or None")
-    return int(start_frame), int(end_frame)
+    if start_frame is not None:
+        start_frame = int(start_frame)
+    if end_frame is not None:
+        end_frame = int(end_frame)
+    return start_frame, end_frame
 
 
 def divide_recording_into_time_chunks(num_frames, chunk_size, padding_size):
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -886,7 +886,7 @@ def cast_start_end_frame(start_frame, end_frame):
         end_frame = end_frame
     else:
         raise ValueError("end_frame must be an int, float (not infinity), or None")
-    return start_frame.astype('int64'), end_frame.astype('int64')
+    return int(start_frame), int(end_frame)
 
 
 def divide_recording_into_time_chunks(num_frames, chunk_size, padding_size):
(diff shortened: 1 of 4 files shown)


spikeextractors/recordingextractor.py
```diff
@@ -496,11 +496,20 @@ def copy_channel_properties(self, recording, channel_ids=None):
         else:
             if isinstance(channel_ids, (int, np.integer)):
                 channel_ids = [channel_ids]
+
```
alejoe91, line 499: This is because otherwise key_properties are duplicated as normal properties as well.

spikeextractors/multirecordingchannelextractor.py
```diff
@@ -35,21 +35,21 @@ def __init__(self, recordings, groups=None):
         # set group information for channels if available
         if groups is not None:
             if len(groups) == len(recordings):
+                group_values = []
                 for i, group in enumerate(groups):
                     recording = recordings[i]
                     channel_ids = recording.get_channel_ids()
-                    recording.set_channel_groups(groups=np.repeat(group, len(channel_ids)), channel_ids=channel_ids)
+                    group_values += [group] * len(channel_ids)
+                self.set_channel_groups(groups=group_values)
```
alejoe91, line 43: This sets groups and locations for the multirecording object directly, without affecting the recordings in the recording list

spikeextractors/extraction_tools.py
```diff
@@ -886,7 +886,7 @@ def cast_start_end_frame(start_frame, end_frame):
         end_frame = end_frame
     else:
         raise ValueError("end_frame must be an int, float (not infinity), or None")
-    return start_frame, end_frame
+    return start_frame.astype('int64'), end_frame.astype('int64')
```
alejoe91, line 889: safety measure (see https://github.com/SpikeInterface/spiketoolkit/issues/409)

spikeextractors/recordingextractor.py
alejoe91: see https://github.com/SpikeInterface/spiketoolkit/issues/411

spikeextractors/extraction_tools.py
samuelgarcia: I don't known underlying reason.
I would prefer a int(start_frame).
Force a cast to a fixed precision is not the spirit I think.

spikeextractors/multirecordingchannelextractor.py
samuelgarcia: Yep better

spikeextractors/multirecordingchannelextractor.py
```diff
@@ -35,21 +35,21 @@ def __init__(self, recordings, groups=None):
...
+        locations = np.empty([0, 2])
         for i, recording in enumerate(recordings):
-            if i == 0:
-                self._key_properties['location'] = recording._key_properties['location']
-            else:
-                self._key_properties['location'] = np.vstack((self._key_properties['location'],
-                                                              recording._key_properties['location']))
+            locations = np.vstack((locations, recording.get_channel_locations()))
+        self.set_channel_locations(locations)
+
```
samuelgarcia, line 52: yep better

spikeextractors/recordingextractor.py
```diff
@@ -496,11 +496,20 @@ def copy_channel_properties(self, recording, channel_ids=None):
...
+            # copy normal properties
             for channel_id in channel_ids:
                 curr_property_names = recording.get_channel_property_names(channel_id=channel_id)
                 for curr_property_name in curr_property_names:
-                    value = recording.get_channel_property(channel_id=channel_id, property_name=curr_property_name)
-                    self.set_channel_property(channel_id=channel_id, property_name=curr_property_name, value=value)
+                    if curr_property_name not in self._key_properties.keys():  # key property
+                        value = recording.get_channel_property(channel_id=channel_id, property_name=curr_property_name)
+                        self.set_channel_property(channel_id=channel_id, property_name=curr_property_name, value=value)
```
samuelgarcia, line 512: This is ok for the moment this should change when implementing the dict of np.array approach

PR #503, opened 2020-11-20
Implement timestamp mechanism

alejoe91:


samuelgarcia:
It is a very very change in the API.
It means that a recording can handle a non continous buffer.
It break all filter like processing and break all sorter also.

This must discuss deeply and collectively.

PR #507, opened 2020-11-26
Extend neuroscope recording to lfp

CodyCBakerPhD:
@bendichter @alejoe91 Extends the `NeuroscopeRecordingExtractor` to be used on the nearly identically formatted `.eeg` files for use in the nwb-conversion-tools LFP data interfaces.

Also minor refactoring of `NwbRecordingExtractor.add_electrical_series`.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -667,7 +667,7 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                             recording=recording,
                             channels_ids=channel_ids
                         ),
-                        iter_axis=1,  # nwb standard is time as zero axis,
+                        iter_axis=1,  # nwb standard is time as zero axis
                         maxshape=(recording.get_num_frames(), recording.get_num_channels())
                     ),
                     compression='gzip'
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -657,7 +657,7 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                 )
             else:
                 def data_generator(recording, channels_ids):
-                    #  generates data chunks for iterator
+                    # generates data chunks for iterator
                     for id in channels_ids:
                         data = recording.get_traces(channel_ids=[id]).flatten()
                         yield data
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -16,7 +16,7 @@ try:
     import pandas as pd
     from pynwb import NWBHDF5IO
     from pynwb import NWBFile
-    from pynwb.ecephys import ElectricalSeries
+    from pynwb.ecephys import ElectricalSeries, LFP
     from pynwb.ecephys import ElectrodeGroup
     from hdmf.data_utils import DataChunkIterator
     from hdmf.backends.hdf5.h5_utils import H5DataIO
@@ -128,6 +128,28 @@ def list_get(l, idx, default):
         return l[idx]
     except IndexError:
         return default
+
+
+def check_module(nwbfile: pynwb.NWBFile, name: str, description: str = None):
+    """
+    Check if processing module exists. If not, create it. Then return module.
+
+    Parameters
+    ----------
+    nwbfile: pynwb.NWBFile
+    name: str
+    description: str | None (optional)
+
+    Returns
+    -------
+    pynwb.module
+    """
+    if name in nwbfile.modules:
+        return nwbfile.modules[name]
+    else:
+        if description is None:
+            description = name
+        return nwbfile.create_processing_module(name, description)
 
 
 class NwbRecordingExtractor(se.RecordingExtractor):
@@ -310,7 +332,7 @@ class NwbRecordingExtractor(se.RecordingExtractor):
             assert isinstance(nwbfile, NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
         defaults = dict(
             name="Device",
-            description="no description"
+            description="Ecephys probe."
         )
         if metadata is None:
             metadata = dict(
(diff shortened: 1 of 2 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -653,36 +653,30 @@ def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metada
...
+                    iter_axis=1,  # nwb standard is time as zero axis
+                    maxshape=(recording.get_num_frames(), recording.get_num_channels())
                 )
+
             if use_timestamps:
                 nwbfile.add_acquisition(
                     ElectricalSeries(
                         name=es_name,
-                        data=ephys_data,
+                        data=H5DataIO(ephys_data, compression="gzip"),
```
alejoe91, line 679: Should we have compression by default?

bendichter:
@CodyCBakerPhD We should not load LFP in the RecordingExtractor.

CodyCBakerPhD:
@bendichter Is that an undesired use within the [SpikeGLX recording extractor](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py) as well? It's essentially the same sort of file structure where the lfp file is at the same location as the raw data, and only has a different suffix. Plus they're essentially the same type of data (traces, channels, including subset channels, metadata) albeit at different sampling rates.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: It was already included by default, I just moved the wrapper down here instead of up around the DataChunkIterators (saves a couple lines of code).

I do think we should always have it by default though; as far as I know there are only beneficial effects. If you disagree we can add an option for disabling it, though, that would be easy.

alejoe91:
@bendichter I think we should extend the `NwbRecordingExtractor.write_recording` with an extra argument (e.g. `lfp=True/False`) so that traces are saved as LFP insted. SI can easily to compute LFP from raw data!

CodyCBakerPhD:
@alejoe91 I'll look into that and discuss with @bendichter what the best approach would be on that. `NwbRecordingExtractor.write_recording` actually would not be a bad place to handle some of this.

bendichter:
I guess my biggest concern was that `RecordingExtractor` objects are for loading data to be sorted, and LFP cannot be sorted so I would be worried that this would be confusing for the user. Handling it in `write_recording` makes sense, though it would be nice to also be able to support writing both raw and LFP data to the same file. Would you just call `write_recording` twice in that case?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -615,97 +639,117 @@ def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metada
...
+            )
+
+        if not write_as_lfp:
+            eseries_kwargs.update(
+                description=metadata['Ecephys']['ElectricalSeries'].get('description', raw_defaults['description'])
+            )
+            nwbfile.add_acquisition(ElectricalSeries(**eseries_kwargs))
+        else:
+            eseries_kwargs.update(
+                description=metadata['Ecephys']['ElectricalSeries'].get('description', lfp_defaults['description'])
```
bendichter, line 750: I think the description in this case should in in Ecephys.LFP

CodyCBakerPhD: Good  catch!

alejoe91:
Ready to merge @CodyCBakerPhD?

CodyCBakerPhD:
@alejoe91 @bendichter Ready for final review, definitely. If it looks good to both of you, then yeah.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -615,97 +639,118 @@ def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metada
...
+        )
+
+        # channels gains - for RecordingExtractor, these are values to cast traces to uV
+        # for nwb, the conversions (gains) cast the data to Volts
+        # To get traces in Volts = data*channel_conversion*conversion
+        gains = np.squeeze([recording.get_channel_gains(channel_ids=[ch])
+                            if 'gain' in recording.get_channel_property_names(channel_id=ch) else 1
+                            for ch in channel_ids])
+        if len(np.unique(gains)) == 1:  # if all gains are equal
+            scalar_conversion = np.unique(gains)[0] * 1e-6
```
alejoe91, line 695: @CodyCBakerPhD is this to convert to uV? I would assume that gains already convert from int16 to uV...I would use gains as they are

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -615,97 +639,118 @@ def add_electrical_series(recording: se.RecordingExtractor, nwbfile=None, metada
...
+            channel_conversion = None
+        else:
+            scalar_conversion = 1.
+            channel_conversion = gains * 1e-6
```
alejoe91, line 699: Same here

alejoe91:
Looks good @CodyCBakerPhD, but I have a comment on the gain conversions

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: It's to convert from uV to V, V is the nwb ElectricalSeries standard unit. If extra conversion values are necessary (in that other line you commented on) for int16 translation, it is  assumed that from spikeextractors, those factors convert it from int16 to uV.

CodyCBakerPhD:
@alejoe91 Hopefully my comment on the gains/conversion factors makes sense; I haven't personally touched that part of the code since it was originally written, only moved it around as needed.

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: Right! There's a minus ;)

alejoe91:
good to go! merging (as soon as tests are done) :)

PR #509, opened 2020-11-27
Add cell explorer sorting extractor

CodyCBakerPhD:
@bendichter Added simple sorting extractor for Cell Explorer format.

@alejoe91 Did I forget to add anything for the tests or imports?

### DIFF
--- a/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
+++ b/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
@@ -26,7 +26,7 @@ class CellExplorerSortingExtractor(SortingExtractor):
     mode = "file"
     installation_mesg = ""
 
-    def __init__(self, spikes_matfile_path: PathType):
+    def __init__(self, spikes_matfile_path: str):
         SortingExtractor.__init__(self)
 
         spikes_matfile_path = Path(spikes_matfile_path)
@@ -47,7 +47,7 @@ class CellExplorerSortingExtractor(SortingExtractor):
         assert len(set(['UID', 'times']) - set(spikes_mat['spikes'].dtype.names)) == 0, \
             "The spikes.cellinfo.mat file must contain a 'spikes' struct with fields 'UID' and 'times'!"
 
-        self._unit_ids = list(spikes_mat['spikes']['UID'][0][0][0])
+        self._unit_ids = list(map(int, spikes_mat['spikes']['UID'][0][0][0]))
         # CellExplorer reports spike times in units seconds; SpikeExtractors uses time units of sampling frames
         self._spiketrains = [
             (np.array([y[0] for y in x]) * sampling_frequency).round().astype(int)
--- a/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
+++ b/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
@@ -26,7 +26,7 @@ class CellExplorerSortingExtractor(SortingExtractor):
     mode = "file"
     installation_mesg = ""
 
-    def __init__(self, spikes_matfile_path: str):
+    def __init__(self, spikes_matfile_path: PathType):
         SortingExtractor.__init__(self)
 
         spikes_matfile_path = Path(spikes_matfile_path)
--- a/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
+++ b/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
@@ -1,13 +1,18 @@
 from spikeextractors import SortingExtractor
 import numpy as np
 from pathlib import Path
-from spikeextractors.extraction_tools import check_valid_unit_id, get_sub_extractors_by_property
+from spikeextractors.extraction_tools import check_valid_unit_id
 from typing import Union
-import re
-from scipy.io import loadmat
+from scipy.io import loadmat, savemat
+
+try:
+    from lxml import etree as et
+
+    HAVE_LXML = True
+except ImportError:
+    HAVE_LXML = False
 
 PathType = Union[str, Path, None]
-DtypeType = Union[str, np.dtype, None]
 
 
 class CellExplorerSortingExtractor(SortingExtractor):
@@ -21,59 +26,19 @@ class CellExplorerSortingExtractor(SortingExtractor):
     """
 
     extractor_name = 'CellExplorerSortingExtractor'
-    installed = True
+    installed = HAVE_LXML
     is_writable = True
     mode = 'custom'
-    installation_mesg = ""
+    installation_mesg = "Please install lxml to use this extractor!"
 
     def __init__(self, matfile_path: PathType):
-
-        # At least one file_path passed
-        if resfile_path is not None:
-            assert clufile_path is not None, 'If passing resfile_path or clufile_path, both are required.'
-            resfile_path = Path(resfile_path)
-            clufile_path = Path(clufile_path)
-            assert resfile_path.is_file() and clufile_path.is_file(), \
-                'The resfile_path ({}) and clufile_path ({}) must be .res and .clu files!'.format(resfile_path,
-                                                                                                  clufile_path)
-
-            assert folder_path is None, 'Pass either a single folder_path location, ' \
-                                        'or a pair of resfile_path and clufile_path. All received.'
-            folder_path_passed = False
-            folder_path = resfile_path.parent
-        else:
-            assert folder_path is not None, 'Either pass resfile_path and clufile_path, or folder_path'
-            folder_path = Path(folder_path)
-            assert folder_path.is_dir(), 'The folder_path must be a directory!'
-
-            res_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.res' in f.suffixes 
-                         and '.temp' not in f.suffixes
-                         and not f.name.endswith('~')
-                         and len(f.suffixes) == 1]
-            clu_files = [f for f in folder_path.iterdir() if f.is_file()
-                         and '.clu' in f.suffixes 
-                         and not f.name.endswith('~')
-                         and len(f.suffixes) == 1]
-
-            assert len(res_files) > 0 or len(clu_files) > 0, \
-                'No .res or .clu files found in the folder_path.'
-            assert len(res_files) == 1 and len(clu_files) == 1, \
-                'NeuroscopeSortingExtractor expects a single pair of .res and .clu files in the folder_path. ' \
-                'For multiple .res and .clu files, use the NeuroscopeMultiSortingExtractor instead.'
-
-            folder_path_passed = True  # flag for setting kwargs for proper dumping
-            resfile_path = res_files[0]
-            clufile_path = clu_files[0]
+        assert HAVE_LXML, self.installation_mesg
+        assert matfile_path.is_file(), f"The matfile_path ({matfile_path}) must exist!"
 
         SortingExtractor.__init__(self)
 
-        res_sorting_name = resfile_path.name[:resfile_path.name.find('.res')]
-        clu_sorting_name = clufile_path.name[:clufile_path.name.find('.clu')]
-
-        assert res_sorting_name == clu_sorting_name, f'The .res and .clu files do not share the same name! ' \
-                                                     f'{res_sorting_name}  -- {clu_sorting_name}'
-
+        matfile_path = Path(matfile_path)
+        folder_path = matfile_path.parent
         xml_files = [f for f in folder_path.iterdir() if f.is_file() if f.suffix == '.xml']
         assert len(xml_files) > 0, 'No .xml file found in the folder.'
         assert len(xml_files) == 1, 'More than one .xml file found in the folder.'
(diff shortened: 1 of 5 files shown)


bendichter:
@CodyCBakerPhD woah, complete with write_sorting? Cool! Does this work for the spec defined [here](https://cellexplorer.org/datastructure/data-structure-and-format/#spikes)?

CodyCBakerPhD:
@bendichter To the extent of the adhered structure of the `UID` and `times` fields, which are all we require here on `spikeextractors`, yes. The `get_metadata` method for the `nwb-conversion-tools` could certainly be extended to include all the fields listed in the remainder of that spec; however, in practice what I have found is the vast majority of those fields are optional and may not be included in the actual `.mat` file, so what I've included thus far is only what I've personally observed to occur in actual data.

bendichter:
@CodyCBakerPhD it looks from the documentation like you might be able to compare between the ts and times fields of the essionName.spikes.cellinfo.mat file to determine the sampling rate. If you do that, maybe you wouldn't need to parse the xml file?

CodyCBakerPhD:
@bendichter That's a good idea; unfortunately, I have yet to observe the `ts` field in any of the `.mat` files for the current datasets reporting spiking unit metadata in this format. There is, however, oftentimes a field unregistered with the documentation merely called `SamplingRate` which I `believe` (but can't 100% guarantee) reports the same value as from the `xml`. I want to say that unregistered field is actually present more often than most of the registered ones, but I would need to triple check that over multiple sessions.

I'll check that the `SamplingRate` field is trustable and if so add checks and retrieval that negate the need for the `xml` parsing (may still have to require `lxml` installed however, in case the field doesn't exist).

alejoe91:
@CodyCBakerPhD imports look good :)

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,111 @@
...
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = np.Inf
+        times = self._spiketrains[self.get_unit_ids().index(unit_id)]
+        inds = np.where((start_frame <= times) & (times < end_frame))
+        return times[inds]
+
+    @staticmethod
+    def write_sorting(sorting: SortingExtractor, save_path: PathType):
```
bendichter, line 88: It looks like this .mat file follows a standard naming structure in the cell explorer format

CodyCBakerPhD: Right you are, added an assertion check to ensure format compliance.

bendichter: Maybe it would be better to have the user just provide the directory path and automatically call this filesorting_id.spikes.cellinfo.mat

CodyCBakerPhD:
@bendichter @alejoe91 On the sampling rate issue, it gets a tad murky between the documentation and reality. 90% of the `spikes.cellinfo.mat` files across four randomly chosen experiments (and about 3 randomly chosen sessions from each) contain a `samplingRate` field that is indeed the same as the rate reported in the Neuroscope `.xml`.

However, since the `.xml` is fundamentally a part of the Neuroscope infrastructure, not the CellExplorer, I'm opting to remove it. Instead, there are other CellExplorer `.mat` files that contain the rate information with 100% reliability over all such files I have observed _in practice_, namely the `.sessionInfo.mat`.

I emphasize _in practice_ because the exact way in which these rates are recorded differs slightly from the online documentation for CellExplorer. But it does seem to be followed very consistently, so that's why I'm trusting it enough to use it here as the primary method of retrieving the sampling frequency.

@alejoe91 Oh, also, I noticed on a second pass that the way CellExplorer records time units differs from Neuroscope in that it actually encodes the time in seconds. I've added some code in the `init` to convert it to effective frames according to sampling frequency, as well as in the `write_sorting` to convert it back appropriately.

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: I did try that at first, however it is the `spikeextractors` standard to always have the second ordered argument be a `save_path` that corresponds to the file location (if "file" is the operational mode of the extractor), and all other arguments optional whereas our third argument of 'filesorting_id' would have to be required, or otherwise automatically set by the stem of the parent folder (but that is also not necessarily the CellExplorer format).

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
+from spikeextractors import SortingExtractor
+import numpy as np
+from pathlib import Path
+from spikeextractors.extraction_tools import check_valid_unit_id
+from typing import Union
+from scipy.io import loadmat, savemat
+
+PathType = Union[str, Path, None]
```
bendichter, line 8:
```suggestion
from typing import Optional
from scipy.io import loadmat, savemat

PathType = Optional[str, Path]
```

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        sorting_id = spikes_matfile_path.name.split(".")[0]
+        session_info_matfile_path = folder_path / f"{sorting_id}.sessionInfo.mat"
+        assert session_info_matfile_path.is_file(), "No sessionInfo.mat file found in the folder!"
+        session_info_mat = loadmat(file_name=str(session_info_matfile_path.absolute()))
+        assert session_info_mat['sessionInfo']['rates'][0][0]['wideband'], "The sesssionInfo.mat file must contain " \
+            "a 'sessionInfo' struct with field 'rates' containing field 'wideband'!"
+        sampling_frequency = float(
+            session_info_mat['sessionInfo']['rates'][0][0]['wideband'][0][0][0][0]
+        )  # careful not to confuse it with the lfpsamplingrate; reported in units Hz
+        self._sampling_frequency = sampling_frequency
```
bendichter, line 44:
```suggestion
        self._sampling_frequency = float(
            session_info_mat['sessionInfo']['rates'][0][0]['wideband'][0][0][0][0]
        )  # careful not to confuse it with the lfpsamplingrate; reported in units Hz
```

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        Path to the sorting_id.spikes.cellinfo.mat file.
+    """
+
+    extractor_name = "CellExplorerSortingExtractor"
+    installed = True
+    is_writable = True
+    mode = "file"
+    installation_mesg = ""
+
+    def __init__(self, spikes_matfile_path: PathType):
```
bendichter, line 29: It seems like since you are reading data from more than one file in this directory you should give the directory path instead of the filepath.

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        session_info_mat = loadmat(file_name=str(session_info_matfile_path.absolute()))
+        assert session_info_mat['sessionInfo']['rates'][0][0]['wideband'], "The sesssionInfo.mat file must contain " \
+            "a 'sessionInfo' struct with field 'rates' containing field 'wideband'!"
+        sampling_frequency = float(
+            session_info_mat['sessionInfo']['rates'][0][0]['wideband'][0][0][0][0]
+        )  # careful not to confuse it with the lfpsamplingrate; reported in units Hz
+        self._sampling_frequency = sampling_frequency
+
+        spikes_mat = loadmat(file_name=str(spikes_matfile_path.absolute()))
+        assert len(set(['UID', 'times']) - set(spikes_mat['spikes'].dtype.names)) == 0, \
```
bendichter, line 47:
```suggestion
        assert np.all(np.isin(['UID', 'times'], names)), \
```
This feels cleaner to me

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+            "The spikes.cellinfo.mat file must contain a 'spikes' struct with fields 'UID' and 'times'!"
+
+        self._unit_ids = list(map(int, spikes_mat['spikes']['UID'][0][0][0]))
```
bendichter, line 50:
```suggestion
        self._unit_ids = np.asarray(spikes_mat['spikes']['UID'][0][0][0], dtype=int)
```

Again, using numpy feels cleaner to me here

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        # CellExplorer reports spike times in units seconds; SpikeExtractors uses time units of sampling frames
+        self._spiketrains = [
+            (np.array([y[0] for y in x]) * sampling_frequency).round().astype(int)
```
bendichter, line 53:
```suggestion
            (np.array([y[0] for y in x]) * sampling_frequency).astype(int)
```

Is the rounding necessary? casting to `int` should do this for you automatically

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        self._kwargs = {'spikes_matfile_path': str(spikes_matfile_path.absolute())}
+
+    def get_unit_ids(self):
+        return list(self._unit_ids)
+
+    @check_valid_unit_id
+    def get_unit_spike_train(self, unit_id, start_frame=None, end_frame=None):
+        start_frame, end_frame = self._cast_start_end_frame(start_frame, end_frame)
+        if start_frame is None:
+            start_frame = 0
```
bendichter, line 66:
```suggestion
```

This shouldn't be necessary

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        if end_frame is None:
+            end_frame = np.Inf
+        times = self._spiketrains[self.get_unit_ids().index(unit_id)]
+        inds = np.where((start_frame <= times) & (times < end_frame))
```
bendichter, line 70: this may be a good place for `np.searchsorted`

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: It is unfortunately; casting to `int` clips the remainder by taking the floor and that in conjunction with the sampling frequency conversion actually produces the occasional value error during testing.

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: While it would totally be doable to swap it to a `folder_path` setup instead of `file_path`, I'm always basing these designs and styles after other extractor types. SpikeGLX and Neuroscope both have this same type of structure where you're reading the actual data from the primary `file_path`, but also pulling metadata from a secondary file in the same folder.

@alejoe91 Which mode do you think would be more appropriate in these types of situations?

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: Definitely cleaner in this case, since there's no naming duplicating that would require unique extraction.

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
bendichter: I see... I'll defer to @alejoe91 on that then

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
bendichter: ok, this is a good solution then. Would you mind adding this in the code comments?

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: Actually, I think it's just better if we remove the None inclusion here so it's clearer that it's non-optional. This was copied over from a structure that supported multiple path options, but we don't have that here.

bendichter:
This is an important extractor because it builds a bring to/from [CellExplorer](https://cellexplorer.org/), an analysis program from the Buzsaki Lab that I think would be useful for many SpikeInterface users. Aside from a few minor style suggestions above, this looks good from my end. @alejoe91 would you mind taking a look at this? In particular, could you take a look at the convo about supplying an input file vs. directory? I'll defer to you for that.

bendichter:
fyi @petersenpeter

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: Sure thing!

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: Agreed

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
CodyCBakerPhD: I'll forego this on this PR since this could be a greater ecosystem wide refactor. Raised [an issue on our fork](https://github.com/catalystneuro/spikeextractors/issues/44) and assigned myself to it, will look into soon.

petersenpeter:
Hi guys

This is great. Let me know if I can be of help. 

Regarding the sampling rate. This is stored in the *.XML, the *.sessionInfo.mat and *.session.mat files. These are "kind of" three interactions of our format. I have never stored the sampling rate with the spikes struct but it seems that some of the datasets have this. You can also derive the sampling rate from waveforms timestamps: spikes.timeWaveform

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -0,0 +1,103 @@
...
+        return times[inds]
```
bendichter, line 72:
```suggestion
        start_frame_index, end_frame_index = np.searchsorted([start_frame, end_frame], times))
        return times[start_frame_index:end_frame_index]
```

CodyCBakerPhD: Straightforward copying of this into the code produces errors in the extractor tests; I'll debug it offer it as a separate ecosystem wide PR in the next couple days. This type of indexing is used in a lot of different extractors so it'll be quite nice standardize.

bendichter: hmm ok

PR #512, opened 2020-11-29
Add neuroscope multirecordingtime extractor

CodyCBakerPhD:
@bendichter @alejoe91 The PeyracheA dataset from the Buzsaki lab reports some raw data broken up by chronological epochs. Considering this the structure used there as standard for the Neuroscope format, this class uses the `MultiRecordingTimeInterface` to  read in these multiple `.dat` files and combine them into a single extractor for conversion purposes. Highly analogous to the `NeuroscopeMultiSorting` extractor.

### DIFF
--- a/spikeextractors/extractorlist.py
+++ b/spikeextractors/extractorlist.py
@@ -26,8 +26,8 @@ from .extractors.neoextractors import (PlexonRecordingExtractor, PlexonSortingEx
                                        NeuralynxRecordingExtractor, NeuralynxSortingExtractor,
                                        BlackrockRecordingExtractor, BlackrockSortingExtractor,
                                        MCSRawRecordingExtractor)
-from .extractors.neuroscopeextractors import NeuroscopeRecordingExtractor, NeuroscopeSortingExtractor,\
-    NeuroscopeMultiSortingExtractor
+from .extractors.neuroscopeextractors import NeuroscopeRecordingExtractor, NeuroscopeMultiRecordingTimeExtractor, \
+    NeuroscopeSortingExtractor, NeuroscopeMultiSortingExtractor
 from .extractors.waveclussortingextractor import WaveClusSortingExtractor
 from .extractors.yassextractors import YassSortingExtractor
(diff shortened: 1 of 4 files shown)


bendichter:
@CodyCBakerPhD Does `NeuroscopeMultiRecordingTimeExtractor` need its own `write_recording`? Maybe it would be more efficient to just modify `NeuroscopeRecordingExtractor.write_recording` instead.

CodyCBakerPhD:
@bendichter It needs it for the extractor tests. And given how it operates over internally set `epochs` in the `RecordingExtractor`, it could be useful for general purposes as well.

I do get the concern since it's the most complicated part of the code, though. It was a similar deal with the `NeuroscopeMultiSortingExtractor` (whose structure this is largely based off of).

bendichter:
Can you just do

```python
def write_recording(*args, **kwargs):
    NeuroscopeRecordingExtractor.write_recording(*args, **kwargs)
````
?

CodyCBakerPhD:
Nope, it has to conform to and replicate the type of multi-file format that the `__init__` expects to read in, which is actually slightly different in that there are multiple `.dat` files but only one `.xml` file. This is actually why the current method doesn't even use `NeuroscopeRecordingExtractor.write_recording` in an iterated fashion; it would spit out a new `.xml` for each `.dat` file we wanted to write.

Of course, if you and @alejoe91 are OK with this particular class not having extractor tests, we can always just remove the need for this `write_recording` method entirely. Though it would allow anyone to take any recording extractor object and write it out in this particular Multi Neuroscope fashion, I kind of doubt anyone would ever actually want to do that.

bendichter:
ok, in that case this looks good to me. @alejoe91 can you look at this when you get a chance?

samuelgarcia:
Hi Ben and Cody.
Note that I plan to refactor the spikeextractor base object to handle directly and more explicitly "multi segment" recording (which  are now handle no explictly by "MultiRecordingTimeExtractor".
This  big change is not for now maybe late in December or in January and will not be merge before some month.
But keep in mind that this code to handle multi segment will certainly need also a refactoring when this change will be effective.

PR #522, opened 2020-12-10
Implemented 'object_properties' in BaseExtractor

alejoe91:
@samuelgarcia @colehurwitz 

This implements `object_properties`. It comes very handy for example to store channel locations in a sorting extractor.

Let's discuss!

### DIFF
--- a/spikeextractors/baseextractor.py
+++ b/spikeextractors/baseextractor.py
@@ -23,7 +23,7 @@ class BaseExtractor:
         self._tmp_folder = None
         self._key_properties = {}
         self._properties = {}
-        self._object_properties = {}
+        self._annotations = {}
         self._memmap_files = []
         self._features = {}
         self._epochs = {}
@@ -88,11 +88,11 @@ class BaseExtractor:
 
         if self.is_dumpable:
             dump_dict = {'class': class_name, 'module': module, 'kwargs': self._kwargs,
-                         'key_properties': self._key_properties, 'object_properties': self._object_properties,
+                         'key_properties': self._key_properties, 'annotations': self._annotations,
                          'version': version, 'dumpable': True}
         else:
             dump_dict = {'class': class_name, 'module': module, 'kwargs': {}, 'key_properties': self._key_properties,
-                         'object_properties': self._object_properties, 'version': imported_module.__version__,
+                         'annotations': self._annotations, 'version': imported_module.__version__,
                          'dumpable': False}
         return dump_dict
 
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Could we call it "annotations" ? It is not a joke.

alejoe91:
> Could we call it "annotations" ? It is not a joke.

ok for me

alejoe91:
@samuelgarcia better?

spikeextractors/baseextractor.py
```diff
@@ -271,11 +273,71 @@ def allocate_array(self, memmap, shape=None, dtype=None, name=None, array=None):
                 arr = np.zeros(shape, dtype=dtype)
         return arr
 
+    def set_annotation(self, annotation_name, value, overwrite=False):
```
samuelgarcia, line 276: in neo we call annotate(self, **kargs):
and this support several annoation.

spikeextractors/baseextractor.py
```diff
@@ -271,11 +273,71 @@ def allocate_array(self, memmap, shape=None, dtype=None, name=None, array=None):
...
+            The data associated with the given property name. Could be many
+            formats as specified by the user
+        '''
+        if annotation_name not in self._annotations.keys():
+            print(f"{annotation_name} is not an annotation")
+            return None
+        else:
+            return deepcopy(self._annotations[annotation_name])
+
+    def get_annotation_names(self):
```
samuelgarcia, line 317: get_annotations_keys()

spikeextractors/baseextractor.py
alejoe91: So the thing I don't like about `**kwargs` is that it prevents using spaces..but I can call it annotate

spikeextractors/baseextractor.py
alejoe91, line 317:
```suggestion
    def get_annotation_keys(self):
```

spikeextractors/baseextractor.py
samuelgarcia: space for what ? for annotation name ?
You can, annotate({'my spaced key': v}) is this is really needed.

alejoe91: I would keep the same signature as it is now

samuelgarcia:
Ok lets merge.

spikeextractors/baseextractor.py
```diff
@@ -273,26 +273,26 @@ def allocate_array(self, memmap, shape=None, dtype=None, name=None, array=None):
                 arr = np.zeros(shape, dtype=dtype)
         return arr
 
-    def set_annotation(self, annotation_name, value, overwrite=False):
+    def annotate(self, annotation_key, value, overwrite=False):
```
samuelgarcia, line 276: This is realy neo friendly :)

PR #524, opened 2020-12-12
Various Nwb improvements

CodyCBakerPhD:
@bendichter @alejoe91 

1. Small fix to accessing `frame_to_time` of the final frame of an epoch; it's not usually an issue elsewhere because the strategy is to read `start_frame:end_frame` for a given epoch... but for the nwb purposes we just extract the two points individually and convert them to time to write to the NWBFile, so trying to access a value right at the end returns an error.

2. Added `overwrite` option to `write_recording` and `write_sorting`

3. Added some default descriptions for unit properties auto-generated by spike interface

4. Fixed default NWBFile values to new standard


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -776,9 +776,12 @@ def add_epochs(recording: se.RecordingExtractor, nwbfile=None,
         for epoch_name in recording.get_epoch_names():
             epoch = recording.get_epoch_info(epoch_name)
             if nwbfile.epochs is None:
+                end_frame = epoch['end_frame']
+                if epoch['end_frame'] == recording.get_num_frames():
+                    end_frame -= 1
                 nwbfile.add_epoch(
                     start_time=recording.frame_to_time(epoch['start_frame']),
-                    stop_time=recording.frame_to_time(epoch['end_frame']),
+                    stop_time=recording.frame_to_time(end_frame),
```
bendichter, line 784: How did we end up with an improper end frame to begin with? Perhaps that is the better place to fix this.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -913,30 +917,32 @@ def write_recording(recording: se.RecordingExtractor, save_path: PathType = None
...
 
         if nwbfile is None:
-            if Path(save_path).is_file():
+            if Path(save_path).is_file() and not overwrite:
                 read_mode = 'r+'
             else:
                 read_mode = 'w'
 
-            with NWBHDF5IO(save_path, mode=read_mode) as io:
+            with NWBHDF5IO(str(Path(save_path).absolute()), mode=read_mode) as io:
```
bendichter, line 925: Why do we need this?

CodyCBakerPhD: NWBHDF5IO doesn't support Path objects, only strings. But `spikeextractors` generally accepts `save_paths` to be Path objects, potentially.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -1232,13 +1245,13 @@ def write_units(sorting: se.SortingExtractor, nwbfile,
...
-            print("Warning: The nwbfile already contains units. "
-                  "These units will not be over-written.")
+        else:
+            warnings.warn("The nwbfile already contains units. These units will not be over-written.")
 
     @staticmethod
     def write_sorting(sorting: se.SortingExtractor, save_path: PathType = None, nwbfile=None,
-                      property_descriptions: dict = None, timestamps: ArrayType = None, **nwbfile_kwargs):
+                      property_descriptions: dict = None, timestamps: ArrayType = None, overwrite: bool = False,
+                      **nwbfile_kwargs):
```
bendichter, line 1254: Need to add overwrite to docstring too

bendichter (changes requested):
A few questions. Looks good overall.

spikeextractors/extractors/nwbextractors/nwbextractors.py
bendichter: OK, but why are we casting as a `Path` and taking `absolute`? `NWBHDF5IO` can write to a relative path, can't it?

CodyCBakerPhD: It's casted since save_path can also optionally be a string to begin with. Suppose we don't need the absolute, though.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Added

spikeextractors/extractors/nwbextractors/nwbextractors.py
bendichter: I still do not understand why you would need to cast it as a `Path`

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: I'll try to completely retrace it; this particular case was discovered via upstream `frame_to_time` conversions being routed through a `SubRecordingExtractor` of a `MultiRecordingTimeExtractor` (hence why we had epochs to begin with) containing the custom `SyntalosRecordingExtractor`. At all the steps, I didn't really notice an improper use from what I could see, maybe @alejoe91 knows more about that.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: We don't if we don't use the absolute anymore, removing

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: Yeah I think that the problem here is that the frames within an epoch are `start_frame <= frames < end_frame`. So in order to get the right timestamps we should always use: `stop_time=recording.frame_to_time(epoch['end_frame'] - 1)`

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -1077,19 +1080,28 @@ def write_units(sorting: se.SortingExtractor, nwbfile,
...
-                    print(f"Warning: description for property {pr} not found in property_descriptions. "
-                          f"Setting description to 'no description'")
+        default_descriptions = dict(
+            isi_violation="Quality metric that measures the ISI violation ratio as a proxy for the purity of the unit.",
+            firing_rate="Number of spikes per unit of time.",
+            template="The extracellular average waveform.",
+            max_channel="The recording channel id with the largest amplitude.",
+            halfwidth="The full-width half maximum of the negative peak computed on the maximum channel.",
+            peak_to_valley="The duration between the negative and the positive peaks computed on the maximum channel.",
+            snr="The signal-to-noise ratio of the unit."
```
alejoe91, line 1090: nice1

PR #525, opened 2020-12-12
first attempt to get GIN testing working

bendichter:
depends on https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/issues/29

* move tests outside of package (they have different dependencies from the package itself)
* move utilities to testing module within package so they can be imported
* add test_gin_repo. I'm not sure if this is working yet. The download take a while
* Add requirements-dev.txt, which specifies the requirements for running the tests


bendichter:
It looks like the CI workflow is not using requirements-dev.txt

CodyCBakerPhD:
True that - given that it's not a GitHub Actions (and that I don't have authorization on this repo), I'll let Alessio know to look into that on the travis end.

I'm just finishing up getting this to work locally to begin with, but I have to push/pull bigger changes between my comp and smaug since the git annex external doesn't want to install for my Windows machine.

alejoe91:
Superseeded by https://github.com/catalystneuro/spikeextractors/tree/add_GIN_tests_cn

PR #526, opened 2020-12-14
NeuroscopeRecordingExtractor gains

CodyCBakerPhD:
@bendichter @alejoe91 Added an extra argument for `NeuroscopeRecordingExtractor` that specifies proper gain values.

### DIFF
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -30,8 +30,8 @@ class NeuroscopeRecordingExtractor(BinDatRecordingExtractor):
     ----------
     file_path : str
         Path to the .dat file to be extracted.
-    gain : float
-        Numerical value that converts the native int dtype to microvolts.
+    gain : float, optional
+        Numerical value that converts the native int dtype to microvolts. Defaults to 1.
     """
 
     extractor_name = "NeuroscopeRecordingExtractor"
@@ -40,7 +40,7 @@ class NeuroscopeRecordingExtractor(BinDatRecordingExtractor):
     mode = "file"
     installation_mesg = "Please install lxml to use this extractor!"
 
-    def __init__(self, file_path: PathType, gain: float):
+    def __init__(self, file_path: PathType, gain: float = 1.0):
         assert HAVE_LXML, self.installation_mesg
         file_path = Path(file_path)
         assert file_path.is_file() and file_path.suffix in [".dat", ".eeg"], \
@@ -70,7 +70,7 @@ class NeuroscopeRecordingExtractor(BinDatRecordingExtractor):
                                           dtype=dtype, numchan=numchan_from_file)
         self.set_channel_gains(channel_ids=list(range(numchan_from_file)), gains=gain)
 
-        self._kwargs = dict(file_path=str(Path(file_path).absolute()), gain=1)
+        self._kwargs = dict(file_path=str(Path(file_path).absolute()), gain=gain)
 
     @staticmethod
     def write_recording(recording: RecordingExtractor, save_path: PathType, dtype: DtypeType = None,
@@ -152,8 +152,8 @@ class NeuroscopeMultiRecordingTimeExtractor(MultiRecordingTimeExtractor):
     ----------
     folder_path : PathType
         Path to the .dat files to be extracted.
-    gain : float
-        Numerical value that converts the native int dtype to microvolts.
+    gain : float, optional
+        Numerical value that converts the native int dtype to microvolts. Defaults to 1.
     """
 
     extractor_name = "NeuroscopeMultiRecordingTimeExtractor"
@@ -162,7 +162,7 @@ class NeuroscopeMultiRecordingTimeExtractor(MultiRecordingTimeExtractor):
     mode = "folder"
     installation_mesg = "Please install lxml to use this extractor!"
 
-    def __init__(self, folder_path: PathType, gain: float):
+    def __init__(self, folder_path: PathType, gain: float = 1.0):
         assert HAVE_LXML, self.installation_mesg
 
         folder_path = Path(folder_path)
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -71,7 +71,7 @@ class NeuroscopeRecordingExtractor(BinDatRecordingExtractor):
                                           dtype=dtype, numchan=numchan_from_file)
 
         if gain is not None:
-            self.set_channel_gains(channel_ids=list(range(numchan_from_file)), gains=gain)
+            self.set_channel_gains(channel_ids=self.get_channel_ids(), gains=gain)
 
         self._kwargs = dict(file_path=str(Path(file_path).absolute()), gain=gain)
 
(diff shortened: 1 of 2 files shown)


spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -29,16 +29,18 @@ class NeuroscopeRecordingExtractor(BinDatRecordingExtractor):
...
+    extractor_name = "NeuroscopeRecordingExtractor"
     installed = HAVE_LXML
     is_writable = True
-    mode = 'file'
-    installation_mesg = 'Please install lxml to use this extractor!'  # error message when not installed
+    mode = "file"
+    installation_mesg = "Please install lxml to use this extractor!"
 
-    def __init__(self, file_path: PathType):
+    def __init__(self, file_path: PathType, gain: float):
```
bendichter, line 43: should have default value of 1.0

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -66,8 +68,9 @@ def __init__(self, file_path: PathType):
 
         BinDatRecordingExtractor.__init__(self, file_path, sampling_frequency=sampling_frequency,
                                           dtype=dtype, numchan=numchan_from_file)
+        self.set_channel_gains(channel_ids=list(range(numchan_from_file)), gains=gain)
 
-        self._kwargs = {'file_path': str(Path(file_path).absolute())}
+        self._kwargs = dict(file_path=str(Path(file_path).absolute()), gain=1)
```
bendichter, line 73: Why `str(Path(file_path)`?

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -157,17 +162,17 @@ class NeuroscopeMultiRecordingTimeExtractor(MultiRecordingTimeExtractor):
     mode = "folder"
     installation_mesg = "Please install lxml to use this extractor!"
 
-    def __init__(self, folder_path: PathType):
+    def __init__(self, folder_path: PathType, gain: float):
```
bendichter, line 165: Should have default of 1.0

bendichter (changes requested):

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: This is actually one case I would advocate having the absolute usage, since this is an object that gets saved and might not be re-loaded from the same relative path.

bendichter: ok

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: I do have concerns about setting this as an optional with a default, since it could lead to the original issue we had. If it could truly be determined automatically, I would go with that but would it not be better to just force the user to think about it?

bendichter: * It's not needed for many common use-cases of a spike sorting pipeline, because it is not needed to do spike sorting
* It changes this API in a breaking way, which will certainly break any code that depends on it
* minor: It is different from the standard API for `RecordingExtractors`

CodyCBakerPhD: Granted, making it a requirement breaks any back-compatibility...

But then again, any previous instances of the use of this extractor wouldn't have recorded the traces in the correct units anyway... so should probably make people aware if they were using it in that capacity?

CodyCBakerPhD: OK, I'll make it optional then. Shall I throw a warning at least if it defaults?

bendichter: @alejoe91, what do you think? Do users of this package generally care about the conversion factor?

alejoe91: I think this should be up to the user. I would leave it as an optional arg with no warning.

alejoe91:
looks good to me @CodyCBakerPhD

bendichter:
@CodyCBakerPhD maybe we could make the gain `NaN` by default? That would make it clear that it is using the default and not to be trusted. Would that mess anything up on your end, @alejoe91 ?

alejoe91:
> @CodyCBakerPhD maybe we could make the gain `NaN` by default? That would make it clear that it is using the default and not to be trusted. Would that mess anything up on your end, @alejoe91 ?

Maybe even `None`, but then also as Cody suggested if we want to notify the user that it's using 1 we could print a warning.

CodyCBakerPhD:
@bendichter @alejoe91 We could set the default to `None`, and only set the gains as a property if it is not `None`. Any attempts to retrieve gains from the extractor without the gains via `get_channel_gains()` would return a default value of 1 (this is used in the `NwbRecordingExtractor.write_recording()`, for instance), but would not actually add the gains as a property, indicating they haven't actually be set.

This compared to actually setting them all to be 1, implying someone and some point actually know that the conversion factor was exactly 1, and set it to that extent.

bendichter:
Why not have `get_channel_gains()` return `None` if unset?

alejoe91:
> Why not have `get_channel_gains()` return `None` if unset?

Because that would break downstream code that uses `get_channel_gains`. I think it's fair to leave 1 if unset

CodyCBakerPhD:
Or actually, I think I was wrong about that defaulting behavior. It looks like the defaulting of 1 is what was built in long ago for conversion factors on the `NwbRecordingExtractor.write_recording()` side, [here](https://github.com/catalystneuro/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L693-L695).

I'm testing it on the toy extractors now - if gain isn't a channel property, trying to access [it should raise an error](https://github.com/catalystneuro/spikeextractors/blob/f20b1219eba9d3330d5d7cd7ce8d8924a255b8c2/spikeextractors/recordingextractor.py#L421-L433) as it isn't a "key" property.

CodyCBakerPhD:
Yeah, can confirm this is the behavior. If you have a `RecordingExtractor` with no gains, trying to call `get_channel_gains()` returns downstream property access error.

Swapped the inclusion behavior here to start off as `None` and only fill in specified.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -67,7 +70,10 @@ def __init__(self, file_path: PathType):
         BinDatRecordingExtractor.__init__(self, file_path, sampling_frequency=sampling_frequency,
                                           dtype=dtype, numchan=numchan_from_file)
 
-        self._kwargs = {'file_path': str(Path(file_path).absolute())}
+        if gain is not None:
+            self.set_channel_gains(channel_ids=list(range(numchan_from_file)), gains=gain)
```
alejoe91, line 74:
```suggestion
            self.set_channel_gains(channel_ids=self.get_channel_ids(), gains=gain)
```

CodyCBakerPhD:
@alejoe91 Any further thoughts on this PR? Working well on my end

PR #530, opened 2020-12-22
added clear properties

colehurwitz:
Added functions for clearing location and group properties. Fixed small inconsistencies and refactored. Added new tests.

I also made it so that if any channel location has a `np.nan` value, then it will not be listed as a property (previous implementation was that if there were any **non**-`np.nan` values, then it would be listed as a property). Thoughts on this @alejoe91 ?

### DIFF
--- a/spikeextractors/recordingextractor.py
+++ b/spikeextractors/recordingextractor.py
@@ -283,7 +283,7 @@ class RecordingExtractor(ABC, BaseExtractor):
         if isinstance(channel_ids, (int, np.integer)):
             channel_ids = [channel_ids]
         #Reset to default locations (NaN)
-        default_locations =  [[np.nan, np.nan, np.nan] for i in range(len(channel_ids))]
+        default_locations =  np.array([[np.nan, np.nan, np.nan] for i in range(len(channel_ids))])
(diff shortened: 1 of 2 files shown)


spikeextractors/recordingextractor.py
```diff
@@ -264,6 +269,23 @@ def get_channel_locations(self, channel_ids=None, locations_2d=True):
...
+        ----------
+        channel_ids: array-like or int
+            The channel ids (ints) for which the locations will be cleared. If None, all channel ids are assumed.
+        '''
+        if channel_ids is None:
+            channel_ids = list(self.get_channel_ids())
+        if isinstance(channel_ids, (int, np.integer)):
+            channel_ids = [channel_ids]
+        #Reset to default locations (NaN)
+        default_locations =  [[np.nan, np.nan, np.nan] for i in range(len(channel_ids))]
```
alejoe91, line 286:
```suggestion
        default_locations =  np.array([[np.nan, np.nan, np.nan] for i in range(len(channel_ids))])
```

alejoe91:
Looks good @colehurwitz :)

PR #531, opened 2020-12-28
small fixes to write methods for nwb recording and sorting

CodyCBakerPhD:
@bendichter @alejoe91 A couple small fixes for things that popped up in the CED pipeline.

1) Apparently, it's possible for the occasional recording extractor to have a sampling frequency that is an integer. Something in pynwb/hdmf wanted it to be a float, so this casts it.

2) For context, max_channel was a property we developed way back for the Yuta set to denote reference electrodes and link them in the spiking units table.

Apparently this same key is automatically set by certain types of returned sorting extractors from `spikesorters` in our pipelines. This isn't an issue in and of itself, _however_ the use of `quick_write` methods allow a possibility that we write a sorting extractor prior to writing any recording information, including an electrode table and hence we cannot complete the link of the electrode table reference.

The solution is simply to not establish the link if that is the case, and warn the user that is the case so that if they do want to include that link, they need to write recording information prior to the sorting extractor with the property set.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -1095,9 +1095,9 @@ class NwbSortingExtractor(se.SortingExtractor):
                 warnings.warn(f"Description for property {pr} not found in property_descriptions. "
                               f"Setting description to 'no description'")
 
+        skip_properties = ["mda_max_channel"]
         if nwbfile.units is None:
             # Check that array properties have the same shape across units
-            skip_properties = []
             property_shapes = dict()
             for pr in all_properties:
                 shapes = []
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -1069,8 +1069,9 @@ class NwbSortingExtractor(se.SortingExtractor):
             sorting: se.SortingExtractor,
             nwbfile,
             property_descriptions: Optional[dict] = None,
-            timestamps: Optional[ArrayType] = None,
-            skip_properties: Optional[List[str]] = None
+            skip_properties: Optional[List[str]] = None,
+            skip_features: Optional[List[str]] = None,
+            timestamps: Optional[ArrayType] = None
         ):
         """Auxilliary function for write_sorting."""
         unit_ids = sorting.get_unit_ids()
@@ -1103,6 +1104,8 @@ class NwbSortingExtractor(se.SortingExtractor):
                 )
         if skip_properties is None:
             skip_properties = []
+        if skip_features is None:
+            skip_features = []
 
         if nwbfile.units is None:
             # Check that array properties have the same shape across units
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Also, per Alessio's suggestion, `mda_max_channel` can simply be ignored whenever auto-generated by `ironclust`.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -1095,9 +1095,9 @@ def write_units(sorting: se.SortingExtractor, nwbfile,
                 warnings.warn(f"Description for property {pr} not found in property_descriptions. "
                               f"Setting description to 'no description'")
 
+        skip_properties = ["mda_max_channel"]
```
bendichter, line 1098: This should be an arg of the function

bendichter (changes requested):

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Easy to do; do you think in this case `mda_max_channel` should be by default ignored or add it in as a specified ignore in the pipelines where it's relevant?

alejoe91: @CodyCBakerPhD @bendichter I think that the `mda_max_channel` is loaded in the `MdaSortingExtractor`. Maybe the best option for now would be to make it optional and have ironclust not loading it by default.

However, we should have a `skip_properties` and `skip_features` as extra arguments

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -1125,14 +1125,16 @@ def write_units(sorting: se.SortingExtractor, nwbfile,
...
-                # Special case of setting max_electrodes requires a table to be
-                # passed to become a dynamic table region
+                # Special case of setting max_electrodes requires a table to be passed to become a dynamic table region
                 if pr not in skip_properties:
                     if pr in ['max_channel', 'max_electrode']:
                         if nwbfile.electrodes is None:
-                            warnings.warn("Attempted to make a custom column for max_channel "
-                                          "or max_electrode, but there are no electrodes to reference! "
-                                          "Column will not be added.")
+                            warnings.warn("Attempted to make a custom column for max_channelor max_electrode, but "
```
alejoe91, line 1132:
```suggestion
                            warnings.warn("Attempted to make a custom column for max_channel or max_electrode, but "
```

CodyCBakerPhD:
@alejoe91 Added and tested. the `skip_properties` and `skip_features` arguments.

Also visually organized the blocks where example extractors are created for the tests, in order to more easily determine which values were included in each extractor; the only actual difference in terms of function is that SX2 now has a sampling frequency. This was mostly motivated by discovery of the final addition below.

@bendichter It is possible for `SortingExtractors` to have a `sampling_frequency` of `None`, meaning it wasn't set at any point. All the actual extractor formats I've dealt with set this from the data so it hasn't come up, but in principle it could occur for a general `SortingExtractor` object.

I've added a hard `ValueError` check for this in the `NwbSortingExtractor` writing functions, since I assume NWB really does require this to get all values in terms of time. Is this appropriate or, in the case where the sampling frequency isn't known, should we simply write the spike times in units of sampling frames (the extractor standard)?

PR #537, opened 2021-01-11
Neuroscope refactoring

CodyCBakerPhD:
@alejoe91 Standardized styling and format for the four extractors, updated docstrings, removed unneeded absolute paths. Very minor changes overall, but makes it much easier to view the changelogs of some upcoming PR's inheriting from this.

Biggest computational change is the removal of `np.loadtxt`, [which is quite slow](https://stackoverflow.com/questions/52232559/numpy-loadtxt-is-way-slower-than-open-readlines), and instead just read the lines directly.


alejoe91:
@CodyCBakerPhD looks good! Can I merge?

CodyCBakerPhD:
@alejoe91 Yep!

PR #538, opened 2021-01-12
fix reader

luiztauffer:
a small fix on NeoBaseRecordingExtractor


samuelgarcia:
Hi @luiztauffer.
Thank you for the fix. Sorry for that stupid error, I think it was me.

About the typing, it is OK for me. But we should have a global choice either type everything either not.
In my mind having all core classes with typping is something good.
Having all extactor with typing is something that could limit contriobutions.

samuelgarcia:
Can I merge ?

PR #541, opened 2021-01-13
Add return_scaled to RecordingExtractor get_traces()

### LINKED ISSUE #540: Multiple signals per channel on BlackrockRecordingExtractor
`BlackrockRecordingExtractor`, which uses `NeoBaseRecordingExtractor`, suffers from the problem (noted elsewhere in the code) of ambiguous channels when there are multiple `ns*` files, containing signals sampled in different rates for the same channels.

I think this could be dealt with in two ways:
1. pass an extra argument to Neo telling it to map only one specific file. Is that a solution simple (and desirable) to be implemented in Neo? 
2. We can map the files indirectly with `BlackrockRecordingExtractor`, given that groups (files) differ in their sampling rate. This is more circumventing the problem and sounds like prone to future problems

@samuelgarcia @alejoe91 ideas?

luiztauffer:
fix #540

### DIFF
--- a/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
+++ b/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
@@ -63,7 +63,7 @@ class BinDatRecordingExtractor(RecordingExtractor):
         return self._sampling_frequency
 
     @check_get_traces_args
-    def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
+    def get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_scaled=True):
         if np.all(channel_ids == self.get_channel_ids()):
             recordings = self._timeseries[:, start_frame:end_frame]
         else:
@@ -77,7 +77,7 @@ class BinDatRecordingExtractor(RecordingExtractor):
             exp_idx = self._dtype.find('int') + 3
             exp = int(self._dtype[exp_idx:])
             recordings = recordings.astype('float32') - 2**(exp - 1)
-        if self._gain is not None:
+        if self._gain is not None and return_scaled:
             recordings = recordings * self._gain
         return recordings
 
--- a/spikeextractors/extractors/mea1kextractors/mea1kextractors.py
+++ b/spikeextractors/extractors/mea1kextractors/mea1kextractors.py
@@ -1,7 +1,7 @@
 from spikeextractors import RecordingExtractor, SortingExtractor
 from pathlib import Path
 import numpy as np
-from spikeextractors.extraction_tools import check_get_traces_args, check_get_ttl_args, check_valid_unit_id
+from spikeextractors.extraction_tools import check_get_traces_args, check_get_ttl_args, check_get_unit_spike_train
 
 try:
     import h5py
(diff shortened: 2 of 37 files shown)


samuelgarcia:
Hi again.
it is not the only kwargs. You should also do something like  

```python
    def __init__(self, filename: PathType, nsx_to_load: Optional[int] = None,
                 block_index: Optional[float] = None, seg_index: Optional[float] = None, **kwargs):
        super().__init__(filename=filename, nsx_to_load=nsx_to_load, 
                         block_index=block_index, seg_index=seg_inde, **kwargsx)
```

Normaly this already done automatically by the NeoBaseRecordingExtractor.
This implementation only make it more explicit.

luiztauffer:
how about now?

luiztauffer:
Somehow related to this, the `get_traces` method returns the traces already scaled to `uV`, when it would be interesting in some cases to have them as `int`and then have the gain applied whenever needed. That helps us save a lot of space.

I added an argument to `NeoBaseRecordingExtractor.get_traces(..., return_scaled=True)` that gives a choice for that, defaults to True (current behavior)

Also, I added channel properties `gain` and `name` at __init__

samuelgarcia:
This gain is a good idea I add this in mind from a long time.

I wanted to do it at base level not only for neo wrapper.

We already have this option in neo, it is super usefull for spike sorting and saving memory bandwith.

OK for this `return_scaled` but it may change a bit in a futur "big refactoring" I have started very recently.

alejoe91:
@luiztauffer still failing but not for your fault :P 

I know why it's failing and I'll fix it over the weekend. It's due to the decorator that we have for handling input arguments to `get_traces`, which somehow it's messing things up.

luiztauffer:
> @luiztauffer still failing but not for your fault :P
> 
> I know why it's failing and I'll fix it over the weekend. It's due to the decorator that we have for handling input arguments to `get_traces`, which somehow it's messing things up.

thanks @alejoe91 
I just added the new arg `return_scaled`, should the decorator be aware of that?

The `get_traces` method works fine in my code though...

alejoe91:
It should deal with it, but now it's base on the base class signature. I'll update it tomorrow to handle all base args plus class-specific kwargs

alejoe91:
@luiztauffer @samuelgarcia @bendichter 

This required a little more extensive fix than expected. I added `return_scaled` in the base `RecordingExtractor` method and propagated to all extractors. I also improved the way decorators handle the arguments and kwargs. 

Let's check that everything looks correct!

Note that this also needed an update to `spiketoolkit`, PR coming soon.

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
```diff
@@ -77,7 +77,7 @@ def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
             exp_idx = self._dtype.find('int') + 3
             exp = int(self._dtype[exp_idx:])
             recordings = recordings.astype('float32') - 2**(exp - 1)
-        if self._gain is not None:
+        if self._gain is not None and return_scaled:
             recordings = recordings * self._gain
```
samuelgarcia, line 81: I gain is None should and return_scaled=True should raise an error

spikeextractors/extractors/mea1kextractors/mea1kextractors.py
```diff
@@ -181,17 +181,18 @@ def get_sampling_frequency(self):
...
-                return recordings[sorted_idx]
+                signals = self._signals[sorted_channel_ids, start_frame:end_frame][sorted_idx]
             else:
-                return (self._signals[np.array(channel_ids), start_frame:end_frame] * self._lsb).astype('float32')
+                signals = self._signals[np.array(channel_ids), start_frame:end_frame]
         else:
-            return (self._signals[np.array(channel_ids), start_frame:end_frame] * self._lsb).astype('float32')
+            signals = self._signals[np.array(channel_ids), start_frame:end_frame]
+        if return_scaled:
+            return (signals * self._lsb).astype('float32')
```
samuelgarcia, line 195:
```
signals = signals.astype('float32')
signals *=  * self._lsb
```

would avoid a copy I think

spikeextractors/extractors/neoextractors/neobaseextractor.py
```diff
@@ -78,28 +78,42 @@ def __init__(self, block_index=None, seg_index=None, **kargs):
...
+            channel_idxs = np.array([list(channel_ids).index(ch) for ch in channel_ids])
+            # and then to uV
+            scaled_traces *= self.additional_gain[:, channel_idxs]
 
-        # fortunatly neo works with (samples, channels) strides
-        # so transpose to spieextractors wolrd
-        scaled_traces = scaled_traces.transpose()
+            # fortunatly neo works with (samples, channels) strides
+            # so transpose to spieextractors wolrd
+            scaled_traces = scaled_traces.transpose()
```
samuelgarcia, line 112: scaled_traces = scaled_traces.transpose() should not be in "if return_scaled:"

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
alejoe91: I think we should do the following. If gain is None in the constructor, we set it to one. Then we multiply by whatever gain we have. I use the bindat a lot (cache) to save filtered data already converted to uV. In that case the scale should be one, bit I think it should be so by default.

alejoe91:
@CodyCBakerPhD merging into the `return_scaled` branch. Let's finish it up on that!

spikeextractors/extractors/neoextractors/neobaseextractor.py
alejoe91: fixed

spikeextractors/extractors/mea1kextractors/mea1kextractors.py
alejoe91: fixed

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
alejoe91: Fixing BinDat on another PR

PR #544, opened 2021-01-16
Add neuroscope waveforms

CodyCBakerPhD:
@alejoe91 As per our discussion, here is the incorporation waveform data from .spk files to the Neuroscope extractor using the current method of encoding features.

@bendichter Will likely be improved in the near future in terms of memory usage.

### DIFF
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -423,7 +423,7 @@ class NeuroscopeSortingExtractor(SortingExtractor):
             n_samples = int(xml_root.find('neuroscope').find('spikes').find('nSamples').text)
             wf = np.memmap(spkfile_path, dtype=dtype)
             n_channels = int(wf.size / (n_spikes * n_samples))
-            wf = wf.reshape(n_spikes, n_samples, n_channels)
+            wf = np.moveaxis(wf.reshape(n_spikes, n_samples, n_channels), 1, -1)
 
             for unit_id in self.get_unit_ids():
                 if gain is not None:
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -311,6 +311,10 @@ class NeuroscopeSortingExtractor(SortingExtractor):
         Optional. Path to the collection of .res and .clu text files. Will auto-detect format.
     keep_mua_units : bool
         Optional. Whether or not to return sorted spikes from multi-unit activity. Defaults to True.
+    spkfile_path : PathType
+        Optional. Path to a particular .spk binary file containing waveform snippets added to the extractor as features.
+    gain : float
+        Optional. If passing a spkfile_path, this value converts the data type of the waveforms to units of microvolts.
     """
 
     extractor_name = "NeuroscopeSortingExtractor"
@@ -325,7 +329,8 @@ class NeuroscopeSortingExtractor(SortingExtractor):
         clufile_path: OptionalPathType = None,
         folder_path: OptionalPathType = None,
         keep_mua_units: bool = True,
-        spkfile_path: OptionalPathType = None
+        spkfile_path: OptionalPathType = None,
+        gain: Optional[float] = None
     ):
         assert HAVE_LXML, self.installation_mesg
         assert not (folder_path is None and resfile_path is None and clufile_path is None), \
@@ -416,6 +421,11 @@ class NeuroscopeSortingExtractor(SortingExtractor):
                 folder_path=None,
                 keep_mua_units=keep_mua_units
             )
+            
+        if spkfile_path is not None:
+            self._kwargs.update(spkfile_path=str(spkfile_path.absolute()))
+        else:
+            self._kwargs.update(spkfile_path=spkfile_path)
 
     def get_unit_ids(self):
         return list(self._unit_ids)
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -421,9 +421,7 @@ class NeuroscopeSortingExtractor(SortingExtractor):
             n_bits = int(xml_root.find('acquisitionSystem').find('nBits').text)
             dtype = f"int{n_bits}"
             n_samples = int(xml_root.find('neuroscope').find('spikes').find('nSamples').text)
-            wf = np.memmap(spkfile_path, dtype=dtype)
-            n_channels = int(wf.size / (n_spikes * n_samples))
-            wf = np.moveaxis(wf.reshape(n_spikes, n_samples, n_channels), 1, -1)
+            wf = np.moveaxis(np.memmap(spkfile_path, dtype=dtype).reshape(n_spikes, n_samples, -1), 1, -1)
 
             for unit_id in self.get_unit_ids():
                 if gain is not None:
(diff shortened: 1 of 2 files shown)


spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -531,7 +566,9 @@ def __init__(
         self,
         folder_path: PathType,
         keep_mua_units: bool = True,
-        exclude_shanks: Optional[list] = None
+        exclude_shanks: Optional[list] = None,
+        write_waveforms: bool = False,
```
alejoe91, line 570: this sounds more as a `load_waveforms` to me

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -519,6 +549,11 @@ class NeuroscopeMultiSortingExtractor(MultiSortingExtractor):
     exclude_shanks : list
         Optional. List of indices to ignore. The set of all possible indices is chosen by default, extracted as the
         final integer of all the .res.%i and .clu.%i pairs.
+    write_waveforms : bool
```
alejoe91, line 552:
```suggestion
    load_waveforms : bool
```

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -572,30 +609,45 @@ def __init__(
...
-                    clufile_path=clufile_path,
-                    keep_mua_units=keep_mua_units
-                )
+            nse_args = dict(
+                resfile_path=folder_path / f"{sorting_name}.res.{shank_id}",
+                clufile_path=folder_path / f"{sorting_name}.clu.{shank_id}",
+                keep_mua_units=keep_mua_units
             )
 
+            if write_waveforms:
```
alejoe91, line 618:
```suggestion
            if load_waveforms:
```

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -572,30 +609,45 @@ def __init__(
...
+
         MultiSortingExtractor.__init__(self, sortings=all_shanks_list_se)
 
         if exclude_shanks_passed:
             self._kwargs = dict(
                 folder_path=str(folder_path.absolute()),
                 keep_mua_units=keep_mua_units,
-                exclude_shanks=exclude_shanks
+                exclude_shanks=exclude_shanks,
+                write_waveforms=write_waveforms,
```
alejoe91, line 641:
```suggestion
                load_waveforms=load_waveforms,
```

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -572,30 +609,45 @@ def __init__(
...
+                gain=gain
             )
         else:
             self._kwargs = dict(
                 folder_path=str(folder_path.absolute()),
                 keep_mua_units=keep_mua_units,
-                exclude_shanks=None
+                exclude_shanks=None,
+                write_waveforms=write_waveforms,
```
alejoe91, line 649:
```suggestion
                load_waveforms=load_waveforms,
```

alejoe91:
@CodyCBakerPhD looks good, but IMO `write_waveforms` should be `load_waveforms`!

CodyCBakerPhD:
@alejoe91 Yeah, that's fair. Changed to `load_waveforms`

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -410,20 +417,43 @@ def __init__(
                 for s_id in self._unit_ids:
                     self._spiketrains.append(res[(clu == s_id + 1).nonzero()])  # from 2,...,clu[0]-1
 
+        if spkfile_path is not None and Path(spkfile_path).is_file():
+            n_bits = int(xml_root.find('acquisitionSystem').find('nBits').text)
+            dtype = f"int{n_bits}"
+            n_samples = int(xml_root.find('neuroscope').find('spikes').find('nSamples').text)
+            wf = np.memmap(spkfile_path, dtype=dtype)
+            n_channels = int(wf.size / (n_spikes * n_samples))
+            wf = np.moveaxis(wf.reshape(n_spikes, n_samples, n_channels), 1, -1)
```
bendichter, line 426:
```suggestion
            wf = np.memmap(spkfile_path, dtype=dtype, shape=(n_spikes, n_samples, -1))
            wf = np.moveaxis(wf, 1, -1)
```

see https://numpy.org/doc/stable/reference/generated/numpy.memmap.shape.html

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -410,20 +417,43 @@ def __init__(
...
                 clufile_path=str(clufile_path.absolute()),
                 folder_path=None,
-                keep_mua_units=keep_mua_units
+                keep_mua_units=keep_mua_units,
+                gain=gain
             )
+        if spkfile_path is not None:
+            self._kwargs.update(spkfile_path=str(spkfile_path.absolute()))
+        else:
+            self._kwargs.update(spkfile_path=spkfile_path)
```
bendichter, line 456: Do we want to set this if it's `None`?

bendichter:
Looks good, @CodyCBakerPhD! I made 2 minor suggestions

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: I almost always use the shape argument on memmap instantiation, but in this case it would be a bit messier then usual for a couple reasons.

1) The memmap `shape` argument doesn't automatically fill dimension values via `-1`, only reshape handles that. 
2) As seen in the code, the `n_spikes` can be easily inferred from the `.clu` file, the `n_samples` can be easily found in the `.xml`; the `n_channels` information is technically within the `.xml` file as well, but would be just a tad more complicated to extract (estimating an additional 2-4 lines of code to do it safely) because these lines of code support both single shank and multi shank channel groups, and the only information we have about which channel group it is comes from. Would also have to debug it on single-pair res/clu .xml, which I haven't done in a while.
3) Even beyond the extra code (2) would requires, there's actually practical reasons to use `-1` instead of reading the true channel groups from the `.xml` - on occasion, I notice that certain channels are outright removed from the `.spk` file, likely because they were considered "bad" at some point in post-processing. As a result, there's a mismatch between the groups reported in the `.xml` and the actual amount of data in the `.spk`.

That said, I have condensed these three lines into one single call, if that looks better.

spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
CodyCBakerPhD: Yeah, dumped `kwargs` should always be set even if their default of `None`. Also present in code lines just above this, but this one is separated due to the fact that the `spkfile_path` argument itself does not interact with the `folder_path_passed` auto-detection scheme, since the very act of passing the `spkfile_path` acts like a boolean check that the user intends to load the waveform data in addition to the sorting data, which is a very optional choice.

bendichter: > dumped kwargs should always be set even if their default of None
why is that?

CodyCBakerPhD: Hmm... not actually sure, one would think that trying to load back a dumped extractor would just pass the kwargs dynamically so optionals would default if not present.

@alejoe91 Is there a more fundamental reason to have optional defaults set for `kwarg` dumping?

PR #547, opened 2021-01-25
return_scaled - bindat, neuroscope, and intan

CodyCBakerPhD:
@alejoe91 Got around to updating this per our discussion

### DIFF
--- a/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
+++ b/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
@@ -42,12 +42,17 @@ class BinDatRecordingExtractor(RecordingExtractor):
 
         if geom is not None:
             self.set_channel_locations(self._geom)
+
         if 'numpy' in str(dtype):
             dtype_str = str(dtype).replace("<class '", "").replace("'>", "")
             # drop 'numpy
             dtype_str = dtype_str.split('.')[1]
         else:
             dtype_str = str(dtype)
+
+        if gain is not None:
+            self.set_channel_gains(channel_ids=self.get_channel_ids(), gains=gain)
+
         self._kwargs = {'file_path': str(Path(file_path).absolute()), 'sampling_frequency': sampling_frequency,
                         'numchan': numchan, 'dtype': dtype_str, 'recording_channels': recording_channels,
                         'time_axis': time_axis, 'geom': geom, 'offset': offset, 'gain': gain,
--- a/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
+++ b/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
@@ -1,33 +1,79 @@
-from spikeextractors import RecordingExtractor
-from spikeextractors.extraction_tools import read_binary, write_to_binary_dat_format, check_get_traces_args
 import shutil
 import numpy as np
 from pathlib import Path
-import os
+from typing import Union, Optional
+
+from spikeextractors import RecordingExtractor
+from spikeextractors.extraction_tools import read_binary, write_to_binary_dat_format, check_get_traces_args
+
+PathType = Union[Path, str]
+DtypeType = Union[str, np.dtype]
+OptionalDtypeType = Optional[DtypeType]
+OptionalArrayType = Optional[Union[np.ndarray, list]]
 
 
 class BinDatRecordingExtractor(RecordingExtractor):
+    """
+    Extracts raw neural recordings from binary .dat files.
+
+    The recording extractor always returns channel IDs starting from 0.
+
+    The recording data will always be returned in the shape of (num_channels, num_frames).
+
+    Parameters
+    ----------
+    file_path : str
+        Path to the .dat file to be extracted.
+    sampling_frequency : float
+        The frequency at which the recording was sampled.
+    numchan : int
+        The total number of channels in the binary file.
+    dtype : dtype
+        The data type used in the binary file.
+    recording_channels : ArrayType , optional
+        Specification that subsets the range of (0, numchan) in extraction from the file.
+    time_axis : int, optional
+        The axis of the data matrix to be considered time (either 0 or 1). Defaults to 0.
+    geom : ArrayType, optional
+        Specification of the location for each chaannel.
+    offset: int
+        Number of bytes in the file to offset by during memmap instantiation.
+    gain : float, optional
+        Numerical value that converts the native dtype to microvolts.
+    is_filtered : bool, optional
+        If the recording is filtered or not. Defaults to False.
+    """
+
     extractor_name = 'BinDatRecordingExtractor'
     has_default_locations = False
-    installed = True  # check at class level if installed or not
+    installed = True
     is_writable = True
-    mode = 'file'
-    installation_mesg = ""  # error message when not installed
+    mode = "file"
+    installation_mesg = ""
 
-    def __init__(self, file_path, sampling_frequency, numchan, dtype, recording_channels=None,
-                 time_axis=0, geom=None, offset=0, gain=None, is_filtered=None):
+    def __init__(
+            self,
+            file_path: PathType,
+            sampling_frequency: float,
+            numchan: int,
+            dtype: DtypeType,
+            recording_channels: OptionalArrayType = None,
+            time_axis: int = 0,
+            geom: OptionalArrayType = None,
+            offset: int = 0,
+            gain: Optional[float] = None,
+            is_filtered: bool = None
+    ):
         RecordingExtractor.__init__(self)
         self._datfile = Path(file_path)
         self._time_axis = time_axis
         self._dtype = str(dtype)
         self._sampling_frequency = float(sampling_frequency)
-        self._gain = gain
         self._numchan = numchan
         self._geom = geom
         self._offset = offset
         self._timeseries = read_binary(self._datfile, numchan, dtype, time_axis, offset)
 
-        # keep track of filter status when dumping
         if is_filtered is not None:
             self.is_filtered = is_filtered
         else:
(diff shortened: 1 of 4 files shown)


spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py
```diff
@@ -61,16 +63,29 @@ def get_sampling_frequency(self):
         return self._fs
 
     @check_get_traces_args
-    def get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_scaled=True, dtype=None):
+    def get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_scaled: Optional[bool] = None,
```
alejoe91, line 66: @CodyCBakerPhD can you make rerurn_scaled not optional and set it to True by default?

CodyCBakerPhD: @alejoe91 I played around with that at first but it makes it more complicated to error check without being able to tell if the user manually specified it as True or False, or if it is in default mode. Also note that it is always 'optional' throughout the repo, since it has a default behavior (True). That behavior is mirrored here, and everything in the default mode acts as though it were True. In order to provide the most robust checking against the potentially conflicting `dtype` argument, it really helps to be able to tell when the user sets it, which we can't tell if we just leave it as `True` and allow `dtype` to pass along with an unspecified `return_scale`.

For example, a user that wanted to try to get the base `int` type returned might call

```
my_intan_re = se.IntanRecordingExtractor(intan_file_path)
my_intan_re.get_traces(dtype='int16')
```

but if `return_scaled=True` by default, it would either have to get overridden by the `dtype` (which might imply that the result is scaled, but also somehow an `int` type), or return an error telling the user they have to manually specify both `return_scaled` and `dtype` at `get_traces`. Likewise they could do

```
my_intan_re = se.IntanRecordingExtractor(intan_file_path, dtype='int16')
my_intan_re.get_traces() or my_intan_re.get_traces(return_scaled=True)
```

which, if `return_scaled=True` by default, would force adherence to that and return float traces, which could be quite confusing.

Basically, the way it works right now is, by detecting which arguments were actually specified by the user, we can both provide a clear path of which arguments override each other based on what the user is trying to do (i.e., if `return_scaled` is specified but `dtype` is not, use the implied behavior of `return_scaled`; if `dtype` is specified but `return_scaled` is not, use the `dtype` behavior). If neither is specified in `get_traces` we use the internal `dtype` set at initialization. If both are passed, and conflict with each other, an informative message is given in this respect.

alejoe91: Thanks @CodyCBakerPhD, but in this case `return_scaled: Optional[bool] = None` means that if the user doesn't pass it, it is `None`. IMO, in order to get `int16`, the user should do:
```
my_intan_re.get_traces(return_scaled=False, dtype='int16')
```

Since `return_scaled` should be `True` by default, I think that:
```
my_intan_re.get_traces(dtype='int16')
```

Should throw an error. In case `return_scaled` is `False` but `dtype` is not set, I would return `uint16`, which is how the data is stored. 

Also, since we have all this options for `get_traces`, I think we can remove `dtype` as an argument to the `__init__`. What do you think?

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
```diff
@@ -42,12 +42,17 @@ def __init__(self, file_path, sampling_frequency, numchan, dtype, recording_chan
...
+
         if 'numpy' in str(dtype):
             dtype_str = str(dtype).replace("<class '", "").replace("'>", "")
             # drop 'numpy
             dtype_str = dtype_str.split('.')[1]
         else:
             dtype_str = str(dtype)
+
+        if gain is not None:
+            self.set_channel_gains(channel_ids=self.get_channel_ids(), gains=gain)
```
alejoe91, line 54: good, but we should also update the `get_traces` to use the `gain` property instead of the internal `_gain` field

spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py
CodyCBakerPhD: > Thanks @CodyCBakerPhD, but in this case return_scaled: Optional[bool] = None means that if the user doesn't pass it, it is None.

Yes - but if you read the code below, if it is `None` there are a couple ways it then defaults based on if `dtype` was specified. If you track the default mode from top-down (including `__init__`), it delivers exactly the same as every other extractor as if `return_scaled` was `True`, but we now have that extra piece of information that it was actually passed by the user and not simply defaulted.

While I'm not against removing `dtype` from `__init__` since it really doesn't do much, just keep in mind that removing that breaks back-compatibility, which I'm really trying to preserve here. Also, if we did that, the default `dtype` at `__init__` would have to be `uint16` in order to preserve the gain and offset information.

Another break in back-compatibility also occurs with what you propose on forcing the pairing of `return_scaled=False` with `dtype='int16'` when, under your version, `return_scaled=True` by default. That break I am against, since I know we already make use of that exact function call at several places in currently supported projects, so it would be quite the annoyance to go back in and correct them all. And if it's annoying for me, who knows how many other downstream users would have their workflows disrupted by it.

alejoe91: I agree that we should keep in mind backcompatibility, but note that `return_scaled` already breaks some stuff in other extractors (e.g. NWB write function now requires it..). I think that SI is still immature enough to properly make these changes that would make the API better as a whole!

CodyCBakerPhD:
@alejoe91 

1) `BinDat` no longer has a `self._gain` attribute, and instead utilizes the channel properties for gains. Also added typing annotations. Also added docstring.

2) For `Intan`, `dtype` is removed from `__init__`, and always adds gain/offset information as channel properties. `dtype` is kept in `get_traces`, but `return_scaled` has precedence over it, up to an attempt to pass `return_scaled=False`, `dtype='float',` which throws an error. If `return_scaled=False` and `dtype` is unspecified it defaults to `uint16`.

3) Also reviewed and fixed the impact this argument has on `NwbRecordingExtractor.write_recording`, which was always fundamentally built to operate on unscaled traces, storing gain information separately from the trace call.

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
```diff
@@ -77,66 +135,96 @@ def get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_
             exp_idx = self._dtype.find('int') + 3
             exp = int(self._dtype[exp_idx:])
             recordings = recordings.astype('float32') - 2**(exp - 1)
-        if self._gain is not None and return_scaled:
-            recordings = recordings * self._gain
+        if 'gain' in self.get_shared_channel_property_names() and return_scaled:
+            recordings = recordings * np.reshape(self.get_channel_gains(channel_ids=channel_ids), (len(channel_ids), 1))
```
alejoe91, line 139: @CodyCBakerPhD a nicer way is the following:
```suggestion
            recordings = recordings * self.get_channel_gains(channel_ids=channel_ids)[:, None]
```

spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py
```diff
@@ -61,16 +77,70 @@ def get_sampling_frequency(self):
...
+            Any of 'float', 'int16', or 'uint16'. return_scaled must be False to return 'int16' or 'uint16'.
+        Returns
+        ----------
+        traces: numpy.ndarray
+            A 2D array that contains all of the traces from each channel.
+            Dimensions are: (num_channels x num_frames)
+        """
+        assert dtype in ['float', 'int16', 'uint16', None], "'dtype' must be either 'float', 'int16', or 'uint16'!"
+
+        if return_scaled:
```
alejoe91, line 127: nice!

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -849,8 +847,7 @@ def add_all_to_nwbfile(recording: se.RecordingExtractor, nwbfile=None,
             nwbfile=nwbfile,
             use_timestamps=use_timestamps,
             metadata=metadata,
-            write_as_lfp=write_as_lfp,
-            return_scaled=return_scaled
```
alejoe91, line 853: Why not leaving the option to use `return_scaled=True`? I agree it should be `False` by default though

alejoe91:
Hey @CodyCBakerPhD 

Almost ready to go! I have a few minor comments :)

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Because the trace access here is fundamentally intended to be False, not just by default. Nwb series objects always behave like the equivalent of `get_traces(..., return_scaled=False)` with `gains` set. The only issue we would encounter here is if a given extractor doesn't support `return_scaled=False` (say, it only records them as pre-converted floats), but if no extractor comes immediately to mind on that we can always wait until the situation arises. Even in that case, you'd simply write the series with `return_scaled=True` and conversion factor of `1e-6`.

spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
CodyCBakerPhD: Ahh, very nice!

luiztauffer:
hi guys, is this any close to be merged?

alejoe91:
@CodyCBakerPhD @luiztauffer Yes merging this to the `return_scaled` branch. Since this requires also the other packages to change, I believe it's going to take a bit more time before we merge to master.

PR #548, opened 2021-01-27
Add n_jobs param tp write recording and remove mda_max_channel property

alejoe91:


alejoe91:
@CodyCBakerPhD this prevents the `mda_max_channel` property from automatically being loaded

PR #552, opened 2021-01-30
Fix issue 546

CodyCBakerPhD:
Fix issue 546

@alejoe91 Per internal discussion, checking if the string equivalent of the channel's group id is in the `nwbfile.electrode_groups` before attempting the last-ditch index search.

Also improved metadata defaulting to allow a partial metadata dictionary to be passed. Also added assertion with more informative message in the case where no electrode table is written by consequence of the `continue` warning in the iteration.

@khl02007 This ought to fix your issue, if you can give this branch a try.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -577,20 +577,11 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                             )
                         )
                     else:
-                        if group_id in range(len(nwbfile.electrode_groups)):
-                            group_name = list(nwbfile.electrode_groups.keys())[group_id]
-                            electrode_kwargs.update(
-                                dict(
-                                    group=nwbfile.electrode_groups[group_name],
-                                    group_name=group_name
-                                )
-                            )
-                        else:
-                            warnings.warn("No metadata was passed specifying the electrode group for "
-                                          f"electrode {channel_id}, and the internal recording channel group was "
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -568,22 +569,32 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None, metadata: dic
...
+                                group_name=str(group_id)
                             )
                         )
                     else:
-                        warnings.warn("No metadata was passed specifying the electrode group for "
-                                      f"electrode {channel_id}, and the internal recording channel group was "
-                                      f"assigned a value ({group_id}) outside the indices of the electrode "
-                                      "groups in the nwbfile! Electrode will not be added.")
-                        continue
+                        if group_id in range(len(nwbfile.electrode_groups)):
```
alejoe91, line 580: Why leaving this condition? 

If channel groups are present and the `NwbRecordingExtractor` is used to write to nwb, it will always go into the previous if (`if str(group_id) in nwbfile.electrode_groups`). In case it gets here, I think we just need to raise a warning. using the `range` might in fact select the wrong electrode group anyways. 

@CodyCBakerPhD what do you think?

CodyCBakerPhD: Removed condition. Though it may prevent the function from writing electrodes in particular cases (passing metadata for groups but not electrodes), this will ensure the information of the linking is more accurate and makes fewer assumptions.

PR #555, opened 2021-02-05
property fixes multichannel

### LINKED ISSUE #553: shared_channel_property_names issue: CacheRecordingExtractor of a MultiRecordingChannelExtractor
Hi Guys, I'm testing the effect of whitening in subgroups of channels following this workflow:

1.  I load the recording and add multiple properties to each channel like channelD, probe_number, etc
2. I apply st.preprocessing.whiten to groups of channels using get_sub_extractors_by_property  (for example by probe number) and then combine the sub_extractors in one recording using MultiRecordingChannelExtractor.
3. Finally, I apply mountainsort with the argument grouping_property equals to 'channelD',  'probe_number' or any other property. 

That works fine but then I tried to use a CacheRecordingExtractor between point 2 and 3. I found that the CacheRecordingExtractor loose all the shared_channel_property_names and only keeps ['group', 'location']. Is it a bug or am I just overusing the spikeinterface classes with the intricate workflow?

colehurwitz:
Fixes #553 

Quick explanation: `MultiRecordingChannelExtractor` used to fetch properties from each internal extractor, as a result, the `._properties` dictionary was empty. As `copy_channel_properties` was sometimes deepcopying the `._properties` dictionary, this would cause properties to not be inherited by the `CacheRecordingExtractor` when wrapping a `MultiRecordingChannelExtractor`.

Fix: `MultiRecordingChannelExtractor` now fills the `._properties` dictionary with all of the internal extractor properties and then does not reference the internal extractor properties anymore. This means that **changes to the multiextractor properties will no longer change the internal extractor properties**. Is this behaviour ok @alejoe91 ? I think it is much more straightforward and less prone to error this way.

I added tests as well.

### DIFF
--- a/spikeextractors/multirecordingchannelextractor.py
+++ b/spikeextractors/multirecordingchannelextractor.py
@@ -39,7 +39,9 @@ class MultiRecordingChannelExtractor(RecordingExtractor):
                 for i, group in enumerate(groups):
                     recording = recordings[i]
                     channel_ids = recording.get_channel_ids()
-                    group_values += [group] * len(channel_ids)
+                    recording_groups = [group] * len(channel_ids)
+                    group_values += recording_groups
+                    recording.set_channel_groups(groups=recording_groups)
                 self.set_channel_groups(groups=group_values)
             else:
                 raise ValueError("recordings and groups must have same length")
@@ -49,6 +51,14 @@ class MultiRecordingChannelExtractor(RecordingExtractor):
         for i, recording in enumerate(recordings):
             locations = np.vstack((locations, recording.get_channel_locations()))
         self.set_channel_locations(locations)
+
+        #set all normal properties
+        for channel_id in self.get_channel_ids():
+            recording = self._recordings[self._channel_map[channel_id]['recording']]
+            channel_id_recording = self._channel_map[channel_id]['channel_id']
+            for property_name in recording.get_channel_property_names(channel_id_recording):
+                value = recording.get_channel_property(channel_id_recording, property_name)
+                self.set_channel_property(channel_id=channel_id, property_name=property_name, value=value)
 
         self._kwargs = {'recordings': [rec.make_serialized_dict() for rec in recordings], 'groups': groups}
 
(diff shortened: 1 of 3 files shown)


colehurwitz:
I also removed the deepcopy behaviour of `copy_channel_properties` when channel_ids=None as skipping over the `set_channel_property` function can cause issues for certain key properties. @alejoe91

alejoe91:
Hi @colehurwitz 

I think that this is ok and it's actually a better behavior! If you have a multichannel or multitime rec extractors, it should have it's own properties on top of the internal extractors. If one wants to retrieve the internal extractor with updated properties, one shoul instantiate a new SubExtractor. Do you agree?

colehurwitz:
Thanks for leaving a comment! Currently, the internal extractors do not have any updated properties by default. However, if someone created a new SubExtractor, it would work like this: SubExtractor <-- MultiExtractor <-- internal extractors. So I think the subextractor would have updated properties because it is working on the multiextractor.

spikeextractors/multirecordingchannelextractor.py
```diff
@@ -39,7 +39,9 @@ def __init__(self, recordings, groups=None):
                 for i, group in enumerate(groups):
                     recording = recordings[i]
                     channel_ids = recording.get_channel_ids()
-                    group_values += [group] * len(channel_ids)
+                    recording_groups = [group] * len(channel_ids)
+                    group_values += recording_groups
+                    recording.set_channel_groups(groups=recording_groups)
```
alejoe91, line 44: @colehurwitz as per our discussion about the internal extractors, I think we can remove this
```suggestion
                   
```

spikeextractors/recordingextractor.py
```diff
@@ -539,25 +539,23 @@ def copy_channel_properties(self, recording, channel_ids=None):
             The list (or single value) of channel_ids for which the properties will be copied
         '''
         if channel_ids is None:
-            self._key_properties = deepcopy(recording._key_properties)
```
alejoe91, line 542: @colehurwitz this is much faster and I don't think it's harmful. In order to fix the issue, I think we can just avoid calling `copy_channel_property` from the `MultiRecordingChannelExtractor` or override the function.

spikeextractors/multirecordingchannelextractor.py
colehurwitz: Agreed!

spikeextractors/recordingextractor.py
colehurwitz: It actually works with the `MultiRecordingChannelExtractor` now since all properties are stored in their respective dictionaries. 

I was more concerned about skipping the `set_channel_property` function since it is important for key properties like location and groups. Is the speed upgrade when doing deepcopys very noticeable? I thought the speed increase would be basically negligible since channel properties are always pretty small.

alejoe91: @colehurwitz  calling the `set_channel_property` for all `key_properties` and `properties` is the same as deepcopying the dictionaries. When you have several properties of >1000 channels the deepcopy is way faster. Can you add this back?

colehurwitz: I am fine with having deepcopy back in the code.

colehurwitz:
Looks good! @ferchaure this should solve your problem!

PR #556, opened 2021-02-07
Extractors for Neuropixels recordings

mhhennig:
Implement 2 extractors to handle Neuropixels Open Ephys data:

### NeuropixelsDatRecordingExtractor

This is based on the BinDatRecordingExtractor, and simplifies reading Neuropixels recordings. If provided it can obtain the channel config from the relevant OpenEphys file (settings.xml), otherwise a default is used (probe tip, 384 channels). Channel locations match the four column, hex configuration.

### OpenEphysNPIXRecordingExtractor

Inherits form `OpenEphysRecordingExtractor` and it allows users to load the "AP" or "LFP" streams. Additionally, ttl events can be retrieved thanks to the open ephys implementaiton


alejoe91:
Hi Matthias, 

Thanks for the PR. I think though that we are missing a few things that should be implemented:

1. Handle multi-stream: allow users to load the ap/lf/or nidq streams
2. Multi segments: in openephys you can have several recordings in the same file. 
3. Load TTL events: we should be able to load TTL signals for synchronization. 

These points are all implemented in pyopenephys and the new PR makes some improvements. I think we can just inherit from the OpenEphysRecordingExtractor and add neuropixels specific args if needed. Also, regarding the name, I think we should call it `OpenEphysNPIXExtractor`, because it uses the OE NPIX source. (Neuropixels is too general and it can be recorded with different systems)

I'll make a PR next week and we can discuss there.

mhhennig:
Yes, but...

1. SI as it is does not handle multi stream anyway in a consistent way, and mainly is for spike sorting (why lfp?),
2. ..again, SI does not deal with segments (yet),
3. and TTLs are not supported either.

Can we please have this one as a plain siple Neuropixels extractor, where a single file continuous.dat and (maybe) settings.xml are available. This will be a very common use case, and I need it as data I get from collabs is sent in this format.

On the naming for this one, how about `NeuropixelsBinExtractor` then, to distinguish from the full OpenEphys capable one?

If the full OpenEphys directory structure is there, then one can use your extractor. By the way, the PR you sent me for pyopenephys fails for the data I have, likely because the directory structure does not fully match expectations.

jsiegle:
I think it could be helpful to have a loading function that takes a .dat file and a settings file (e.g., settings.xml from Open Ephys or [.meta file](https://github.com/billkarsh/SpikeGLX/blob/gh-pages/Support/Metadata_3B2.md) from SpikeGLX). However, as Alessio says, this will preclude you from loading LFP data, TTLs, etc. But maybe that's ok for the sake of simplicity.

jsiegle:
On the Open Ephys side, the `SELECTIONSTATE` within each processor actually refers to something different from the selected electrodes in a Neuropixels probe. For the most recent version of the Neuropixels plugin (last ~3 months), this info is saved in the `EDITOR` section.

To load it, first you have to filter the processors by name (just selecting the first one doesn't guarantee you'll get the Neuropixels plugin):

```python
neuropix_source = [proc for proc in settings['SIGNALCHAIN']['PROCESSOR'] 
                                  if proc['@pluginName'] == 'Neuropix-PXI']
```

Then, info about the selected "bank" for each channel is found in the `CHANNELS` field of each `EDITOR/NP_PROBE` item:

```python
selected_electrodes = []

for ch_idx, bank in enumerate(neuropix_source[0]['EDITOR']['NP_PROBE'][0]['CHANNELS'].values()):
    
    selected_electrodes.append(int(bank) * 384 + ch_idx)
```
The electrode index can go up to 960, even though there are only 384 data channels.

Also note that there can be multiple probes recorded simultaneously, so there needs to be some way to disambiguate which probe the .dat file comes from.

Here's a [settings.xml](https://github.com/SpikeInterface/spikeextractors/files/5953128/settings.xml.zip) file with 3 probes in it that I've been using for testing.

spikeextractors/extractorlist.py
```diff
@@ -34,6 +34,7 @@
 from .extractors.alfsortingextractor import ALFSortingExtractor
 from .extractors.cedextractors import CEDRecordingExtractor
 from .extractors.cellexplorersortingextractor import CellExplorerSortingExtractor
+from .extractors.neuropixelsrecordingextractor import NeuropixelsRecordingExtractor
```
alejoe91, line 37: @mhhennig still note convinced about the name..what about `NeuropixelsDatRecordingExtractor`? At least this makes it clear it needs a dat file

spikeextractors/extractors/neuropixelsrecordingextractor/neuropixelsrecordingextractor.py
```diff
@@ -0,0 +1,122 @@
+from spikeextractors import RecordingExtractor
+from spikeextractors.extraction_tools import read_binary, write_to_binary_dat_format, check_get_traces_args
+import shutil
+import numpy as np
+from pathlib import Path
+import os
+import xmltodict
+import warnings
+
+class NeuropixelsRecordingExtractor(RecordingExtractor):
```
alejoe91, line 10: Why not inheriting directly from the `BinDatRecordingExtractor`?

spikeextractors/extractors/neuropixelsrecordingextractor/neuropixelsrecordingextractor.py
```diff
@@ -0,0 +1,122 @@
...
+        else:
+            channel_idxs = np.array([self.get_channel_ids().index(ch) for ch in channel_ids])
+            if np.all(np.diff(channel_idxs) == 1):
+                recordings = self._timeseries[channel_idxs[0]:channel_idxs[0]+len(channel_idxs), start_frame:end_frame]
+            else:
+                # This block of the execution will return the data as an array, not a memmap
+                recordings = self._timeseries[channel_idxs, start_frame:end_frame]
+        if self._gain is not None:
+            recordings = recordings * self._gain
+        return recordings
```
alejoe91, line 122: If you inherit from the BinDat, all this cane be removed :)

alejoe91:
@jsiegle can you test both new extractors? I tested with the 1s file and they both work :)

alejoe91:
merging this

PR #562, opened 2021-02-09
Update SC extractors

alejoe91:
@yger can you check this?

- the `SpykingCircusRecordingExtractor` can now load from NUMPY and RAW_BINARY format

- the `SpykingCircusSortingExtractor` can load a specific curated result file (using e.g. the SC GUI) and it can additionally load the templates


samuelgarcia:
@alejoe91 : does SC sorters use now raw binary ?
One super big reason SC was not using binary is because SC is writting into the source "raw" file for the filetered version.
This is super super dangerous for end user because after SC the original file is not the original one.
If you use several sorters from the same binary file normally no one make a copy axcept SC because SC write the fileter signals into the orinial file.

alejoe91:
This is just to open existing datasets @samuelgarcia (not sorted via SI)

samuelgarcia:
ok

PR #566, opened 2021-02-10
changes to OpenEphysExtractor

luiztauffer:
- [x] have the `pyopenephys.File()` available as an attribute, so NWB data interfaces might have access to some fields that are not necessarily relevant to to the RecordingExtractor (e.g. Date created, acquisition system name)
- [x] get traces in int16 (default) or float (uV)
- [x] set channel gains


spikeextractors/extractors/openephysextractors/openephysextractors.py
```diff
@@ -19,15 +19,20 @@ class OpenEphysRecordingExtractor(RecordingExtractor):
     mode = 'folder'
     installation_mesg = "To use the OpenEphys extractor, install pyopenephys: \n\n pip install pyopenephys\n\n"  # error message when not installed
 
-    def __init__(self, folder_path, *, experiment_id=0, recording_id=0, dtype='float'):
+    def __init__(self, folder_path, *, experiment_id=0, recording_id=0, dtype='int16'):
```
alejoe91, line 22: Hi @luiztauffer 

I'd keep `float` as default, but add a `dtype` arg to the `get_traces` to allow for dynamically select the dtype.

Note that these changes will be soon superseeded by the `return_scaled` arg.

spikeextractors/extractors/openephysextractors/openephysextractors.py
```diff
@@ -40,9 +45,9 @@ def get_sampling_frequency(self):
 
     @check_get_traces_args
     def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
-        if self._dtype == 'int16':
+        if self._dtype == 'int16':   # Returns traces as int16
             return self._recording.analog_signals[0].signal[channel_ids, start_frame:end_frame]
-        elif self._dtype == 'float':
+        elif self._dtype == 'float':   # Returns traces as uV
             return self._recording.analog_signals[0].signal[channel_ids, start_frame:end_frame] * \
                    self._recording.analog_signals[0].gain
```
alejoe91, line 52:
```
def get_traces(self, channel_ids=None, start_frame=None, end_frame=None, dtype=None):
        if dtype is not None:
             assert dtype in ["int16", "float"], "'dtype' can be 'int16' or 'float'"
        else:
             dtype = self._dtype
        if dtype == 'int16':   # Returns traces as int16
            return self._recording.analog_signals[0].signal[channel_ids, start_frame:end_frame]
        elif dtype == 'float':   # Returns traces as uV
            return self._recording.analog_signals[0].signal[channel_ids, start_frame:end_frame] * \
                   self._recording.analog_signals[0].gain
```

alejoe91 (changes requested):

luiztauffer:
@alejoe91 it seems like there's no use for `dtype` on `__init__()`, I think we could just remove it from there.
I can replace already the logic on `get_traces` to accept return_scaled arg instead of dtype.
what do you think?

alejoe91:
@luiztauffer the `dtype`/`return_scaled` changes are already there in the `return_scaled` branch. (https://github.com/SpikeInterface/spikeextractors/blob/return_scaled/spikeextractors/extractors/openephysextractors/openephysextractors.py#L39-L45) 
 
I guess that the only change would be then to expose the `fileobj`. Is it ok if I point this PR to the `return_scaled` branch and we merge there?

luiztauffer:
@alejoe91 sure, we can merge it there! But there are other changes beside fileobj, for instance the if condition is inverted in this code you shared, and gain handling there does not work
but I think they will be solved if we merge these branch there

alejoe91:
@luiztauffer do we still need this after merging https://github.com/SpikeInterface/spikeextractors/pull/572?

luiztauffer:
@alejoe91 no, its the same branch, Im closing it

Issue #564, opened 2021-02-10
ValueError: channel_ids and locations must have same length


gmaggi:
Hello,

I got two issues with the master extractor:

1. spikeextractors/recordingextractor.py", line 237, in set_channel_locations
    raise ValueError("channel_ids and locations must have same length")
ValueError: channel_ids and locations must have same length

is this because of the inputs  that I am passing? In a previous version of spikeextractor (0.9.0), this error did not pop up

2.- there is a type here: https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py#L25 (nameor)

Thanks!

alejoe91:
Hi Giuliano,

How do you get the error? Can you paste your code?

Thanks for finding the typo! Fixing now :) note that we are udating our tests to use example files so we expect that these kind of bugs will be easier to spot in the future.

Cheers
Alessio

gmaggi:
Thanks for the prompt response!!
the code is:

```
recordings=[]
for inputf in inputFiles:
     recording = se.SpikeGLXRecordingExtractor(inputf)
     recordings.append(recording)

subrecording = se.MultiRecordingTimeExtractor(recordings)
```

Below the full error:

`Traceback (most recent call last):
  File "/export/opt/computation_packages/bin/spike_interface_pipeline.py", line 305, in <module>
    recording = run_extractor()
  File "/export/opt/computation_packages/bin/spike_interface_pipeline.py", line 183, in run_extractor
    recording = se.SpikeGLXRecordingExtractor(inputf)
  File "/export/opt/anaconda2/envs/spikeinterface0.11.0_10.02.2021/lib/python3.7/site-packages/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py", line 85, in __init__
    self.set_channel_locations(locations)
  File "/export/opt/anaconda2/envs/spikeinterface0.11.0_10.02.2021/lib/python3.7/site-packages/spikeextractors/recordingextractor.py", line 237, in set_channel_locations
    raise ValueError("channel_ids and locations must have same length")
ValueError: channel_ids and locations must have same length`

alejoe91:
@gmaggi apparently it found a different number of locations than number of channels, and that triggers (correctly) the Error. Does it happen for all recordings? 

Do you mind sharing the recording that raises the error with me? You can send me an email and we can arrange it

[...]

Issue #561, opened 2021-02-09
SpikeGLXRecordingExtractor seems to be saving the wrong channel coordinates


m-col:
Hello!

I was just going through the x, y coordinates of the channels in my neuropixels recording as output from spikeinterface was something looked odd. A search through the neuropixels slack group gave me a short discussion about the problem I had come across (if you don't have access, I can copy the relevant parts): https://neuropixelsgroup.slack.com/archives/C93JDLKJP/p1596032993180600

Long story short, it seems that `SpikeGLXRecordingExtractor` is saving the coordinates as:
```
0 0
21 0
0 20
21 20
...
```
When they should look like:
```
27	0
59	0
11	20
43	20
27	40
...
```
From what I can tell the issue is [here](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py#L150) with the two pitch values.

Has this been brought up before? Do you think it would be a simple matter of changing these values? I'm thinking I'd have to re-run my spike sorting in case the sorter makes different results based on the updated coordinates.

alejoe91:
Hi @m-col 

I don't have access to the slack channel. Could you please share the relevant part? 

The pitch + ShankMap mechanism has been there since the beginning of the extractor. You are right though! It seems that they always have 2 values, while it should be 4. In the files I tested, I thought that the configuration was set to cover the entire probe but using only two columns. 

Can you paste the `~snsShankMap` entry of the `ap.meta` file here?

alejoe91:
BTW, if it's only a matter of the pitch, I added them as an additional arg and corrected the x_pitch (to 16) in this PR:
https://github.com/SpikeInterface/spikeextractors/compare/npix_pitch?expand=1

![image](https://user-images.githubusercontent.com/17097257/107397719-33036f80-6aff-11eb-9339-d8d67ea87e6c.png)

m-col:
> I don't have access to the slack channel. Could you please share the relevant part?

Reading it again I think I summarised all of the main parts (the rest is mentioning SpikeGLX says something different and asking what the truth it). This image was posted which appears to match what you mention in the PR:

![image](https://user-images.githubusercontent.com/44536113/107634071-e46de680-6c60-11eb-9f69-c461d1a63d26.png)

> Can you paste the ~snsShankMap entry of the ap.meta file here?

```
[...]

PR #569, opened 2021-02-11
Fix SpikeGLX locations and channel names

alejoe91:
Fixes https://github.com/SpikeInterface/spikeextractors/issues/564 and https://github.com/SpikeInterface/spikeextractors/issues/561

https://github.com/SpikeInterface/spikeextractors/issues/564
The problem was that the conditions to define AP, LF, or NIDQ recordings were based on the assumptions that `imec0` is in the file name. However, some file names can also only have `imec`

https://github.com/SpikeInterface/spikeextractors/issues/561
The staggered configuration needed to be added 'manually'. Basically every other row will be shifted by x_pitch // 2. The x_pitch has been updated to 32 (16 * 2), as this is the actual distance of the electrodes on the same row.


alejoe91:
@m-col this should fix https://github.com/SpikeInterface/spikeextractors/issues/561 as well

I think that the stagger in the locations needs to be added 'manually'. Basically every other row will be shifted by `x_pitch // 2`. Note that the `x_pitch` has been updated to 32 (16 * 2), as this is the actual distance of the electrodes on the same row.

The channel map looks like this:

image

And the locations are:
```
array([[16.,  0.],
       [48.,  0.],
       [ 0., 20.],
       [32., 20.],
       [16., 40.],
       [48., 40.],
       [ 0., 60.],
       [32., 60.],
       [16., 80.],
       [48., 80.]]
       ...)
```

which is inline with what you mentioned in the issue.Sorry about this issue and thank you for bringing this up!

m-col:
Looks great, thank you for the fix!

PR #571, opened 2021-02-12
return_scaled

CodyCBakerPhD:
@alejoe91 @luiztauffer Getting the draft PR up for all the big combined changes into master related to adding the `return_scaled` argument into `get_traces`.

I personally need to test out the changes to the Nwb writer on actual data - it would be great if the GIN tests were operational in that respect too, for this branch.


alejoe91:
@CodyCBakerPhD could you check the tests? It seems to be caused by the NWB extractor.

BTW, I agree we could intergate the github actions here..or maybe merge the other PR first.

luiztauffer:
the error occurs because we are creating electrode_groups defined at the metadata but these do not correspond necessarily to the groups names extracted by spikeinterface. We could just ignore and use the groups names given by spikeinterface, but this is a problem if we wanna give the user the chance of renaming the electrodes_groups

However, we should be able to provide the basic metadata with get_metadata (which would include the electrode_groups names) and use this metadata (or a user-modified version of it) as input to write_recording.

PR #572, opened 2021-02-16
Openephys

luiztauffer:
bring changes for openephys extractor


luiztauffer:
@alejoe91 that fixed the test error on #571 , should we merge this?

PR #584, opened 2021-02-25
Modified assertion check to accept all imecX.lf or imecX.ap file as input

### LINKED ISSUE #583: handle imec1.ap file
Hello,

One of the spikeinterface-users in my institution is using two probes, every dataset contains a imec0.ap.bin and imec1.ap.bin.

Would it be possible to handle imec1 too?. She currently gets an assert-error from : 

https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py#L38

Thanks,
Giuliano.

MarineChap:
Accept  now
- imec.ap.bin
- imec.lf.bin
- imecX.lf.bin      with X the recording number
- imecX.ap.bin

close #583 

@gmaggi  Do you think you could checkout on this branch on the cluster please ? I want to verify that the pipeline is not breaking later with the imec1 file and also Hanna needs to send her jobs as soon as possible. 

@alejoe91 , @samuelgarcia    should i add some tests to keep track of the allowed filenames ?

### DIFF
--- a/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py
+++ b/spikeextractors/extractors/spikeglxrecordingextractor/spikeglxrecordingextractor.py
@@ -3,7 +3,7 @@ from .readSGLX import readMeta, SampRate, makeMemMapRaw, GainCorrectIM, GainCorr
 import numpy as np
 from pathlib import Path
 from spikeextractors.extraction_tools import check_get_traces_args, check_get_ttl_args
-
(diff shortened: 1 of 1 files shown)


samuelgarcia:
looks good.

MarineChap:
I can confirm now, it is all working as wanted.

MarineChap:
So, when do you think you will able to merge it ?

Issue #573, opened 2021-02-16
Setting timestamps for recording extractors


khl02007:
It would be nice to be able to associate an array of timestamps with a RecordingExtractor. This could then be inherited by the SortingExtractor and used to convert the spike times from samples to seconds. I deal with recordings that are irregularly sampled (i.e. some samples are missing) but each sample is given a timestamp (unit: seconds). So being able to access the timestamps directly from the extractor would be helpful in aligning neural data with other data streams. In that case `frame_to_time` should just look up the timestamp from the array rather than computing based on sample rate.

alejoe91:
Hi @khl02007 

I think that the best option would be to:

- have a `_timestamps` field both for the `RecordingExtractor` and `SortingExtractor`
- have a `set_timestamps` function to allow users to set this externally
- have the `time_to_frame` and `frame_to_time` use this fields internally (I would also add those methods to the `SortingExtractor` for completeness).
- spike sorters should copy the timestamps to the `SortingExtractor`

Note that this will require changes to the `Sub` and `Multi` extractors. Moreover, I think that the `_timestamps` should also be among the variables that are dumped to pickle files, otherwise the information is lost.

A possible pitfall of this approach is that many processing steps in the pipeline assume regularly sampled signals (e.g. filters, spike sorting itself, etc.). While I believe that slight misalignments or a few missing frames won't change the processing too much, it'd still be somehow incorrect. On the other hand, I think that this feature should be available and it should be up to the user to decide whether to use it or not. :)

@khl02007 since you are quite familiar with SI, do you want to give it a try? Maybe make a draft PR and we can discuss over it.

Cheers
Alessio

PR #586, opened 2021-02-27
Added timestamps

alejoe91:
Addresses: https://github.com/SpikeInterface/spikeextractors/issues/573

- adds a `_timestamps` field to the `BaseExtractor`
- adds a `set_timestamps` function to `RecordingExtractor` and `SortingExtractor` (different arg checks)
- `_timestamps` are saved when dumping to pickle
- adds the `time_to_frame` and `frame_to_time` to sorting extractor
- moved epochs mechanism to `BaseExtractor`

- [x] Add tests for timestamps (especially multi-sub extractors)
- [x] In spikesorters: copy epochs and timestamps from recording to sorting


alejoe91:
@samuelgarcia FYI

samuelgarcia:
I did not have time for feedback...

I do it anyway:
  * it should be explicit somewhere in thd doc that this "timestamps" is a time vector in second.
  * I am not super happy with the naming "timestamps", sometimes timestamps is used for units that are not seconds.
  * If we add this times vector we should also add the t_start for continuous sampling situtations. for the frame_to_time.
  * timstamps have to be float64, we need a checker for this.
  * I am not happy with this epoch included at the base level. For me epoch should be a sperated object.
    This should be in a separted PR and discussed.

PR #591, opened 2021-03-04
Base scaling mechanism

alejoe91:
Moves the scaling mechanism to the base `RecordingExtractor` by scaling traces based on `gains` and `offsets`

- traces are scaled in the `check_get_traces_args` decorator, which makes it automatic to all extractors
- `gain` and `offset` are now key properties
- the `CacheRecordingExtractor` now handles both `return_scaled` options
- GIN tests now include dumping and caching

- [ ] Fix saving of NwbRecordingExtractor when `write_scaled=True`. In this case, gains and offsets should not be written to the file, as traces are assumed to be already in uV.


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -620,6 +627,7 @@ def add_electrodes(recording: se.RecordingExtractor, nwbfile=None, metadata: dic
                     description=descr
                 )
 
+    # TODO: do not save gain and offset when write_scaled is True!
```
alejoe91, line 630: @CodyCBakerPhD can you look into this?

spikeextractors/extraction_tools.py
```diff
@@ -788,10 +793,17 @@ def corrected_args(recording, channel_ids=None, start_frame=None, end_frame=None
             warnings.warn("The recording extractor does not have unscaled traces. Returning scaled traces")
             return_scaled = True
 
-        get_traces_correct_arg = func(recording, channel_ids=channel_ids, start_frame=start_frame, end_frame=end_frame,
-                                      return_scaled=return_scaled, **kwargs)
+        traces = func(recording, channel_ids=channel_ids, start_frame=start_frame, end_frame=end_frame,
+                      return_scaled=return_scaled, **kwargs)
+
+        # scaling
+        if recording.has_unscaled and return_scaled:
```
CodyCBakerPhD, line 800: @alejoe91 When/where is the attribute `recording.has_unscaled` set?

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: So, as I look at it especially with respect to the treatment of things like Intan whose return_unscaled is by default now uint16, I'll also need to re-fix the `write_scaled=False` case, too.

spikeextractors/extraction_tools.py
alejoe91: It's defined at the class level (like the `installed`)

spikeextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: I think that is working! But when `write_scaled` is Tru gain and offset should not be stored in the electrode table

CodyCBakerPhD: Again, we really would prefer to avoid writing the offset at all. Ignore I even mentioned the previous solution to that. The `add_electrical_series` should operate on signed int types, or floats. If something returns an unsigned type, we should correct it using the scaled offset.

alejoe91: That's exactly what the gain and offset do right now. return_scaled is True by default, but at least this gives the opportunity to save raw raw data.

CodyCBakerPhD: OK help me out here - I'm walking through the logic of getting traces from an IntanRecordingExtractor: https://github.com/SpikeInterface/spikeextractors/blob/6ebc754511632ac0d134775dbc9b1fb9231888b5/spikeextractors/extractors/intanrecordingextractor/intanrecordingextractor.py#L80-L130

I want my traces to return as `int16`, with `recording.get_channel_gains` returning the conversion factors for each channel. This is the best mapping to the Nwb ElectricalSeries.

But as I look at the decorator: https://github.com/SpikeInterface/spikeextractors/blob/6ebc754511632ac0d134775dbc9b1fb9231888b5/spikeextractors/extraction_tools.py#L762-L807

I can only set `return_scaled` to `True` (returns series as microvolt floats) or `False` (returns the `uint16`).

alejoe91: So, with this update intan can return either uint16 (raw) or float (scaled). The int16 version was a partial conversion that we made because we didn't have the offset mechanism on place

CodyCBakerPhD: Right, because we didn't have the offset mechanism in place on spikeextractors or in the NWB schema. You have it in spikeextractors now but the NWB side will take longer. So I'm saying for the add_electrical_series, I'll have to do that partial conversion thing as the default behavior when `write_scaled` is `False` and `recording.has_unscaled` is `True`. 

Because otherwise we can't tell without trying to infer the dtype, and we could end up writing certain types of raw data to the acquisition without all of the information to naturally convert it to volts.

CodyCBakerPhD:
Aight trying to fix this now, hopefully the CI likes it, if not I'll do specific debugging.

But basically, we always want to set a conversion factor in NWB simply because spikeextractors works in microVolts and we work in Volts. So even in the simplest case, or if write_scaled=True, we still want conversion=1e-6.

I've also added a NotImplementedError for if the coercion of unsigned data (non-zero offsets) isn't possible, notifying the user to simply write data as float since the Nwb schema doesn't officially recognize offsets just yet. I'll try to get the wheels turning on that though, but it very likely won't match the timescale for these big changes.

CodyCBakerPhD:
@alejoe91 is the CI running for this PR?

spikeextractors/extractors/cedextractors/cedrecordingextractor.py
```diff
@@ -91,7 +91,7 @@ def channel_names(self):
 
     @check_get_traces_args
     def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
```
alejoe91, line 93: add `return_scaled`

CodyCBakerPhD:
@alejoe91 Looks good to me!

PR #596, opened 2021-03-10
fix #595

### LINKED ISSUE #595: Repeated unit properties raise an error when writing sorting to nwb
When saving sorting to NWB file, some properties might be repeated and this gives an error. 

image

In this example max_channels will be added twice and this piece of code will raise an error:

https://github.com/SpikeInterface/spikeextractors/blob/3f49cc28a29688116380440794695535313e9361/spikeextractors/extractors/nwbextractors/nwbextractors.py#L63-L66

I'll open a PR solving this

luiztauffer:
fix #595

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -1334,6 +1334,7 @@ class NwbSortingExtractor(se.SortingExtractor):
                     nspks_list = [sp for sp in nspikes.values()]
                     spikes_index = np.cumsum(nspks_list).astype('int64')
                     if ft in nwbfile.units:  # If property already exists, skip it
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -1333,6 +1333,8 @@ def write_units(
                     flatten_vals = [item for sublist in values for item in sublist]
                     nspks_list = [sp for sp in nspikes.values()]
                     spikes_index = np.cumsum(nspks_list).astype('int64')
+                    if ft in nwbfile.units:  # If property already exists, skip it
+                        continue
```
alejoe91, line 1337: @luiztauffer maybe add a warning?

PR #600, opened 2021-03-11
fix issue 587

### LINKED ISSUE #587: 'No acquisitions found in the .nwb file.'
I am trying to follow the tutorial you have online on SpikeInterface. I am trying to use “NwbRecordingExtractor” to load my raw data which is on nwb format. I am attaching the error message that I keep getting. I am also attaching a print of my nwb file. 

Capture
Capture2

CodyCBakerPhD:
fix #587 

Part (a) of that issue dealt with auto-population of the electrical series name, which is now properly fixed.
Part (b) is the odd error that occurs, for example, if the electrode ids in the nwbfile.electrodes table were not added in increasing order - the original direct indexing of the electrical series region (which are indexes of ids in the electrodes table) through the table caused an error through h5py and is now fixed by list compression passing through a single index at a time.
Part (c) is unfortunately the fact that the electrode ids weren't unique to begin with, making it an invalid NWBFile. Reaching out to try to fix in the independent software, but [a PR fix has been proposed on PyNWB](https://github.com/NeurodataWithoutBorders/pynwb/pull/1344), at least.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -6,12 +6,12 @@ import numpy as np
 import distutils.version
 from typing import Union, List, Optional
 import warnings
-import pandas as pd
 
 import spikeextractors as se
 from spikeextractors.extraction_tools import check_get_traces_args, check_get_unit_spike_train
 
 try:
+    import pandas as pd
     import pynwb
     from pynwb import NWBHDF5IO
     from pynwb import NWBFile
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -6,13 +6,13 @@
 import distutils.version
 from typing import Union, List, Optional
 import warnings
+import pandas as pd
```
alejoe91, line 9: Can you keep pandas in the `try`? It's not installed by default )https://github.com/SpikeInterface/spikeextractors/blob/master/setup.py#L20-L24)

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -322,7 +338,7 @@ def get_num_frames(self):
         return self.num_frames
 
     def get_channel_ids(self):
-        return self.channel_ids.tolist()
+        return self.channel_ids
```
alejoe91, line 341: `channel_ids` are lists by convention in SI. Are they already a list?

PR #602, opened 2021-03-16
Yass: zero-spike unit fix

catubc:
Added code to return empty arrays for zero-spike units.


alejoe91:
Why would you return empty spike trains? I think it's better to filter out the unit ids directly. Also why did you change the init? Can you remove the changes to other files other than the yasssortingextractor.py?

catubc:
Hey
So, re: empty spike trains, it turns out that yass was returning a single spike cluster which was not properly indexed by the yassextractor. I fixed that locally, but it should also be fixed in yass to not allow such small units to be returned.

Re: other file changes, I reverted (sorry about that, I'm working from an older forked directory and I think it had some garbage in it).

Let me know if anything else is required.

spikeextractors/extractors/yassextractors/yassextractors.py
```diff
@@ -71,19 +71,21 @@ def get_unit_spike_train(self, unit_id, start_frame=None, end_frame=None):
 
         if self.spike_train is None:
             self.spike_train = np.load(self.fname_spike_train)
-            
+        
         # find unit id spike times
-        idx = np.where(self.spike_train[:,1]==unit_id)
-        spike_times = self.spike_train[idx,0].squeeze()
+        idx = np.where(spike_train[:,1]==unit_id)[0]
```
alejoe91, line 76: @catubc `spike_train` is not defined anywhere...you only define `self.spike_train`

alejoe91:
@catubc see my comment.

You should also rename:

- folder `yasssortingextractor` --> `yassextractors`
- file `yasssortingextractor.py` --> `yassextractors.py`

Maybe it's easier if you checkout master, pull, make a new branch with the changes, make a new PR. If that's ok with you I'll close this.

catubc:
Thanks, I'll circle back soon.
C

PR #604, opened 2021-03-19
Neoextractors channel ids patch

sbuergers:
@alejoe91 
I only changed 1 thing: Forced chan_ids to be np.ndarray of integers. In python-neo it is now by default a np.ndarray of 'U64's, i.e. strings. See here for the changes to python-neo: https://github.com/NeuralEnsemble/python-neo/pull/949

### DIFF
--- a/spikeextractors/extractors/neoextractors/neobaseextractor.py
+++ b/spikeextractors/extractors/neoextractors/neobaseextractor.py
@@ -119,10 +119,16 @@ class NeoBaseRecordingExtractor(RecordingExtractor, _NeoBaseExtractor):
         # there is no garranty that ids/names are unique on some formats
         channel_idxs = [self.get_channel_ids().index(ch) for ch in channel_ids]
         neo_chan_ids = self._neo_chan_ids[channel_idxs]
-        raw_traces = self.neo_reader.get_analogsignal_chunk(block_index=self.block_index, seg_index=self.seg_index,
-                                                            i_start=start_frame, i_stop=end_frame,
-                                                            channel_indexes=None, channel_names=None,
-                                                            stream_index=0, channel_ids=neo_chan_ids)
+        if self.after_v10:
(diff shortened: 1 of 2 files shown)


spikeextractors/extractors/neoextractors/neobaseextractor.py
```diff
@@ -129,6 +129,10 @@ def get_sampling_frequency(self):
 
     def get_channel_ids(self):
         chan_ids = self.neo_reader.header['signal_channels']['id']
+
+        # force chan_ids to be int (new neo default=str)
+        chan_ids = np.array([np.int(i) for i in chan_ids])
```
alejoe91, line 134: @sbuergers NEO now allows any string to be a channel id. I think that this would fail if the channel name is for example `'channel #0'`. I would just make the channel_ids `np.arange(len(chan_ids))` for the time being. 

Note that in the new SI refractor (https://github.com/SpikeInterface/spikeinterface/pull/88) we will allow channel ids to be int or strings.

alejoe91 (changes requested):

alejoe91 (approved):

alejoe91:
@samuelgarcia since now NEO allows for str ids, this PR keeps a copy of the NEO channel ids and redefines (if unique) the channel ids as `np.arange(len(neo_channel_ids))`. The `get_traces` retrieves the correct neo channel ids, that are passed to the `get_analogsignal_chunk` function. 

Does it look fine to you?

samuelgarcia:
What I would do here is to try to convert the array to int.
If it works keep natural ids if not then use this ```list(np.arange(len(self._neo_chan_ids)))```

alejoe91:
> What I would do here is to try to convert the array to int.
> If it works keep natural ids if not then use this `list(np.arange(len(self._neo_chan_ids)))`

With a `try`?

samuelgarcia:
Yep with a try.

PR #607, opened 2021-03-22
allow user-defined metadata when saving to nwb from non-nwb extractor

luiztauffer:
allow user-defined metadata when saving to nwb from non-nwb extractor

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -856,7 +856,7 @@ class NwbRecordingExtractor(se.RecordingExtractor):
             metadata: dict = None,
             write_as_lfp: bool = False,
             write_scaled: bool = False
-    ):
+        ):
         """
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
@luiztauffer Good catch on the dictionary update there

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -856,7 +856,7 @@ def add_all_to_nwbfile(
             metadata: dict = None,
             write_as_lfp: bool = False,
             write_scaled: bool = False
-    ):
+        ):
```
CodyCBakerPhD, line 859: However, though its not a big deal (just something I would change back the next time I do something in this repo) - the indentation level you change here is an E125 violation in PEP8.

luiztauffer: yes, there's something weird in my VScode that makes me do this every time, I'll fix it

PR #612, opened 2021-04-02
Nwbmetadata fixes

alejoe91:
Fixes some metadata handling in NWB write_recording


CodyCBakerPhD:
@alejoe91 Out of curiosity, what led you to start using the metadata functionality in this  way? Keep in mind that while we designed it to be user passable here independently of NWB-conversion-tools, the structure is so rich and complex oftentimes that we really only expect the people developing the `get_metadata` functions within a specific format interface to ever have to deal with it directly.

The changes look fine, **but** to safely test the changes overall, I really should just add some testing infrastructure for the metadata specifically since any minor deviation from what is listed in the docstrings can cause these types of issues.

In general I don't see a problem right off the bat with having `electrode_group` device linkage be automated if only the name of the device in `metadata` is changed with only one device included there.

PR #616, opened 2021-04-06
Fix gain MaxWell extractors

sronchi:


alejoe91:
Thanks @sronchi !! Merging!

Issue #618, opened 2021-04-06
[DOC] - Out of order argument descriptions


TomDonoghue:
I was exploring using SpikeInterface, and using some demos, and I noticed that some functions have the arguments described in the wrong order, which is a bit annoying for doing quick checks. For example `get_traces` and `get_snippets` in `spikeextractors/spikeextractors/recordingextractor.py` have this issue. 

If this is a mistake, I'm happy to do a quick sweep, and submit a PR to update this and any others I find. I just wanted to check first whether this is something to be fixed - I'm not yet super familiar with the module, and so it could be that there is some reason for these orderings that I am unfamiliar with.

alejoe91:
Hi @TomDonoghue 

Thank you for spotting this! Sure if you have time, it would be awesome if you can make a PR to fix it :) it is definitely not intended

PR #619, opened 2021-04-09
[DOC] - Update docstrings.

TomDonoghue:
This PR responds to #618, which was about some out of order arguments in docstrings. As I did a quick sweep to check for occurrences of this, I also did some suggested tweaks and fixes to the docstrings. 

Most updates are organizational / aesthetic, but note that the arguments `channel_maps` and `channel_groups` in `extraction_tools.load_probe_file` and `recordingextractor.load_probe_file` didn't previously have descriptions, so I added some based on what I could infer from the code. These probably want checking for accuracy.

### DIFF
--- a/spikeextractors/extraction_tools.py
+++ b/spikeextractors/extraction_tools.py
@@ -507,7 +507,7 @@ def write_to_h5_dataset_format(recording, dataset_path, save_path=None, file_han
 
 
 def get_sub_extractors_by_property(extractor, property_name, return_property_list=False):
-    """Returns a list of SubRecordingExtractors from the Extractor based on the given
+    """Returns a list of SubExtractors from the Extractor based on the given
     property_name (e.g. group)
 
     Parameters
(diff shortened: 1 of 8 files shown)


spikeextractors/extraction_tools.py
```diff
@@ -502,11 +507,13 @@ def write_to_h5_dataset_format(recording, dataset_path, save_path=None, file_han
 
 
 def get_sub_extractors_by_property(extractor, property_name, return_property_list=False):
-    """Returns a list of SubRecordingExtractors from this RecordingExtractor based on the given
+    """Returns a list of SubRecordingExtractors from the Extractor based on the given
```
alejoe91, line 510:
```suggestion
    """Returns a list of SubExtractors from the Extractor based on the given
```

alejoe91 (changes requested):
Thanks @TomDonoghue for this! Highly appreciated!

I only have a small comment

alejoe91 (approved):

PR #621, opened 2021-04-19
NwbRecordingExtractor.write_recording extended options

luiztauffer:
Extend options when writing to nwb. 
See https://github.com/catalystneuro/nwb-conversion-tools/pull/172

Allows `se.NwbRecordingExtractor.write_recording()` to choose between three options for type of electrical series:
- `raw` will save it in acquisition
- `processed` will save it as FilteredEphys
- `lfp` will save it as LFP

If no nwb metadata is passed, default values are used. 
If nwb metadata with many electrical_series keys is passed, the arg `es_key` should indicate which specific metadata key corresponds to the data being written.

### DIFF
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -711,22 +711,7 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                 description="Processed data",
                 comments="Generated from SpikeInterface::NwbRecordingExtractor"
             )
-        elif write_as == 'lfp':
-            eseries_kwargs = dict(
-                name="ElectricalSeries_lfp",
-                description="Processed data - LFP",
-                comments="Generated from SpikeInterface::NwbRecordingExtractor"
-            )
-
-        # If user passed metadata info, overwrite defaults
-        if es_key is not None and metadata is not None:
-            assert es_key in metadata.get('Ecephys', dict()).keys(), f"Metadata dictionary does not contain key '{es_key}'"
-            eseries_kwargs.update(metadata['Ecephys'][es_key])
-
-        # Check for existing names and processing modules in nwbfile
-        if write_as == 'raw':
-            assert eseries_kwargs['name'] not in nwbfile.acquisition, f"Raw ElectricalSeries '{es_name}' is already written in the NWBFile!"
-        elif write_as == 'processed':
+            # Check for existing processing module and data interface
             ecephys_mod = check_module(
                 nwbfile=nwbfile,
                 name='ecephys',
@@ -735,6 +720,12 @@ class NwbRecordingExtractor(se.RecordingExtractor):
             if 'Processed' not in ecephys_mod.data_interfaces:
                 ecephys_mod.add_data_interface(FilteredEphys(name='Processed'))
         elif write_as == 'lfp':
+            eseries_kwargs = dict(
+                name="ElectricalSeries_lfp",
+                description="Processed data - LFP",
+                comments="Generated from SpikeInterface::NwbRecordingExtractor"
+            )
+            # Check for existing processing module and data interface
             ecephys_mod = check_module(
                 nwbfile=nwbfile,
                 name='ecephys',
@@ -742,6 +733,22 @@ class NwbRecordingExtractor(se.RecordingExtractor):
             )
             if 'LFP' not in ecephys_mod.data_interfaces:
                 ecephys_mod.add_data_interface(LFP(name='LFP'))
+
+        # If user passed metadata info, overwrite defaults
+        if es_key is not None and metadata is not None:
+            assert es_key in metadata.get('Ecephys', dict()).keys(), f"Metadata dictionary does not contain key '{es_key}'"
+            eseries_kwargs.update(metadata['Ecephys'][es_key])
+
+        # Check for existing names in nwbfile
+        if write_as == 'raw':
+            assert eseries_kwargs['name'] not in nwbfile.acquisition, \
+                f"Raw ElectricalSeries '{eseries_kwargs['name']}' is already written in the NWBFile!"
+        elif write_as == 'processed':
+            assert eseries_kwargs['name'] not in nwbfile.processing['ecephys'].data_interfaces['Processed'].electrical_series, \
+                f"Processed ElectricalSeries '{eseries_kwargs['name']}' is already written in the NWBFile!"
+        elif write_as == 'lfp':
+            assert eseries_kwargs['name'] not in nwbfile.processing['ecephys'].data_interfaces['LFP'].electrical_series, \
+                f"LFP ElectricalSeries '{eseries_kwargs['name']}' is already written in the NWBFile!"
 
         # Electrodes table region
         channel_ids = recording.get_channel_ids()
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -736,8 +736,8 @@ class NwbRecordingExtractor(se.RecordingExtractor):
                 ecephys_mod.add(LFP(name='LFP'))
 
         # If user passed metadata info, overwrite defaults
-        if es_key is not None and metadata is not None:
-            assert es_key in metadata.get('Ecephys', dict()).keys(), f"Metadata dictionary does not contain key '{es_key}'"
+        if metadata is not None and 'Ecephys' in metadata and es_key is not None:
+            assert es_key in metadata['Ecephys'], f"metadata['Ecephys'] dictionary does not contain key '{es_key}'"
             eseries_kwargs.update(metadata['Ecephys'][es_key])
 
         # Check for existing names in nwbfile
--- a/spikeextractors/extractors/nwbextractors/nwbextractors.py
+++ b/spikeextractors/extractors/nwbextractors/nwbextractors.py
@@ -15,7 +15,7 @@ try:
     import pynwb
     from pynwb import NWBHDF5IO
     from pynwb import NWBFile
-    from pynwb.ecephys import ElectricalSeries, LFP
+    from pynwb.ecephys import ElectricalSeries, FilteredEphys, LFP
     from pynwb.ecephys import ElectrodeGroup
     from hdmf.data_utils import DataChunkIterator
     from hdmf.backends.hdf5.h5_utils import H5DataIO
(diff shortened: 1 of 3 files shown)


spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -676,8 +676,10 @@ def add_electrical_series(
         use_times: bool (optional, defaults to False)
             If True, the times are saved to the nwb file using recording.frame_to_time(). If False (defualut),
             the sampling rate is used.
-        write_as_lfp: bool (optional, defaults to False)
-            If True, writes the traces under a processing LFP module in the NWBFile instead of acquisition.
+        write_as: str (optional, defaults to 'raw')
+            Options: 'raw', 'lfp' or 'processed'
```
CodyCBakerPhD, line 680: Can you expand on this, letting the user know what precisely it accomplishes?

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -686,61 +688,66 @@ def add_electrical_series(
...
+        elif write_as == 'processed':
+            eseries_kwargs = dict(
+                name="ElectricalSeries_processed",
+                description="Processed data",
+                comments="Generated from SpikeInterface::NwbRecordingExtractor"
+            )
+        elif write_as == 'lfp':
+            eseries_kwargs = dict(
+                name="ElectricalSeries_lfp",
+                description="Processed data - LFP",
```
CodyCBakerPhD, line 714: Can you set this to its previous value?

luiztauffer: what previous value?

spikeextractors/extractors/nwbextractors/nwbextractors.py
luiztauffer: sure!

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -686,61 +688,66 @@ def add_electrical_series(
...
+                comments="Generated from SpikeInterface::NwbRecordingExtractor"
+            )
+
+        # If user passed metadata info, overwrite defaults
+        if es_key is not None and metadata is not None:
+            assert es_key in metadata.get('Ecephys', dict()).keys(), f"Metadata dictionary does not contain key '{es_key}'"
```
CodyCBakerPhD, line 720:
```suggestion
            assert es_key in metadata.get('Ecephys', dict()), f"Metadata dictionary does not contain key '{es_key}'"
```

You don't need to specify `.keys()` when asking if a key is in a dictionary.

spikeextractors/extractors/nwbextractors/nwbextractors.py
luiztauffer: done

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -686,61 +688,66 @@ def add_electrical_series(
...
+            eseries_kwargs.update(metadata['Ecephys'][es_key])
```
CodyCBakerPhD, line 724: Above, you use a safe `.get()` to see if the es_key is in the `metadata['Ecephys']`, but you don't check that `Ecephys` itself is in the metadata. If it is not (and the safe `.get()` returns empty) then this will throw an error.

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -686,61 +688,66 @@ def add_electrical_series(
...
+                name='ecephys',
+                description="Intermediate data from extracellular electrophysiology recordings, e.g., LFP."
+            )
+            if 'Processed' not in ecephys_mod.data_interfaces:
+                ecephys_mod.add_data_interface(FilteredEphys(name='Processed'))
+        elif write_as == 'lfp':
+            ecephys_mod = check_module(
+                nwbfile=nwbfile,
+                name='ecephys',
+                description="Intermediate data from extracellular electrophysiology recordings, e.g., LFP."
```
CodyCBakerPhD, line 741: Might as well combine these blocks with the default population above

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -686,61 +688,66 @@ def add_electrical_series(
...
             ecephys_mod = check_module(
-                nwbfile,
-                'ecephys',
-                "Intermediate data from extracellular electrophysiology recordings, e.g., LFP."
+                nwbfile=nwbfile,
+                name='ecephys',
+                description="Intermediate data from extracellular electrophysiology recordings, e.g., LFP."
+            )
+            if 'Processed' not in ecephys_mod.data_interfaces:
+                ecephys_mod.add_data_interface(FilteredEphys(name='Processed'))
```
CodyCBakerPhD, line 736:
```suggestion
                ecephys_mod.add(FilteredEphys(name='Processed'))
```

spikeextractors/extractors/nwbextractors/nwbextractors.py
luiztauffer: if 'Ecephys' is not in metadata, `metadata.get('Ecephys', dict())` will return an empty dict

spikeextractors/extractors/nwbextractors/nwbextractors.py
luiztauffer: why this?
see https://pynwb.readthedocs.io/en/stable/pynwb.base.html#pynwb.base.ProcessingModule.add_data_interface

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: https://github.com/SpikeInterface/spikeextractors/pull/621/files#diff-6dae9d8400e85905edd7719a4e2c19d9665c6a124cb08bda08cd09a15780ca85L699

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: I see that, but `metadata['Ecephys'][es_key]` will not.

spikeextractors/extractors/nwbextractors/nwbextractors.py
CodyCBakerPhD: Simpler. We haven't used `add_data_interface` in quite a while.

spikeextractors/extractors/nwbextractors/nwbextractors.py
luiztauffer: these default values shouldn't affect anything Also, they can be overwritten by user's metadata if wanted.

spikeextractors/extractors/nwbextractors/nwbextractors.py
luiztauffer: I share the feeling, but if you look at it, we should check the possibility of the `name` coming from the user's metadata already existing in the file, and we can only test it after updating

CodyCBakerPhD: That applies to https://github.com/luiztauffer/spikeextractors/blob/nwb_write_options/spikeextractors/extractors/nwbextractors/nwbextractors.py#L727-L728, sure, but not to the others. From the earliest point that we check the `write_as` values, we know if we need to check the processing modules. Saves us a few lines moving that stuff above.

luiztauffer: moved, but then I noticed we were not testing for existing name when lfp/processed. added that as well

spikeextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -686,61 +691,66 @@ def add_electrical_series(
...
+            eseries_kwargs = dict(
+                name="ElectricalSeries_lfp",
+                description="Processed data - LFP",
+                comments="Generated from SpikeInterface::NwbRecordingExtractor"
+            )
+
+        # If user passed metadata info, overwrite defaults
+        if es_key is not None and metadata is not None:
+            assert es_key in metadata.get('Ecephys', dict()), f"Metadata dictionary does not contain key '{es_key}'"
+            eseries_kwargs.update(metadata['Ecephys'][es_key])
```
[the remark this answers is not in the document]
luiztauffer: done

luiztauffer:
@CodyCBakerPhD thanks for the comments! All the points are covered now

CodyCBakerPhD:
@luiztauffer Looking good - probably want to wait until the GIN tests are back up and running though to make sure nothing gets broken. I'd also consider adding tests to the framework make sure these `write_as` options are actually resulting in the correct operations, especially w.r.t. round-trip

luiztauffer:
@CodyCBakerPhD done with the testing. There's not much extra round trip to be done though, since `NwbRecordingExtractor` reads only from acquisition. All I did is to make sure adding new traces on the file work without errors and that it won't corrupt the acquisition round trip

how mature are the GIN tests? If they’re still under development, no reason to hold back this PR, the default behavior is preserved here. 
I don’t see the usage of the previous `write_as_lfp` in the gin tests. If you were planning to use it, this is the only thing that changed, and it’s a better way now.

CodyCBakerPhD:
> since `NwbRecordingExtractor` reads only from acquisition.

Ohhh good point, I overlooked that when I originally added the `write_as_lfp` option! This should be expanded to include all our write options since they are indeed `RecordingExtractor` objects. But let's leave that for a subsequent PR since this one is a decent size atm.

For tests I mostly wanted the metadata handling to be checked, and that's what you've added so it should be good. GIN is just for extra safety.

luiztauffer:
@CodyCBakerPhD @alejoe91 shall we merge it?

CodyCBakerPhD:
I can't see if the branch is passing the GIN tests, even when I go to the actions tab - @alejoe91 do you know if it's passing? As soon as it is, merging is fine by me

alejoe91:
Seems to be failing :( https://github.com/luiztauffer/spikeextractors/actions/runs/770732151

@luiztauffer can you merge with master? We made some changes to the github action and we fixed the GIN tests.

luiztauffer:
@alejoe91 locally the gin tests are passing... I'm not sure why it is failing here

luiztauffer:
ok, it is passing here as well

alejoe91 (approved):

PR #623, opened 2021-04-21
Debugging GIN tests

CodyCBakerPhD:
@alejoe91 This will fix the GIN test issues with git annex. However, it creates a side issue with the Mac part of the matrix testing b/c the git-annex isn't supported by condaforge for that system I guess. Any ideas for fixing that?

PR #624, opened 2021-04-21
BF?: do configure git in the python-package.yaml CI

yarikoptic:
If works - an alternative to #623 (which would pin git-annex version)

this is just a wild guess/attempt.


alejoe91:
Thanks @yarikoptic 

@CodyCBakerPhD I think that this is a better approach and it should solve the mac issue :)

yarikoptic:
Damn, it seemed to work: https://github.com/SpikeInterface/spikeextractors/actions/runs/773303483 . Eh

alejoe91:
Yeah! Thanks again :)

PR #627, opened 2021-04-28
some fixes for write nwb

luiztauffer:


CodyCBakerPhD:
Please give a detailed list of your problems with these functions, and detailed justification for how your fixes solve them.

luiztauffer:
@CodyCBakerPhD there were wrong keys for pynwb objects and other small fixes (e.g. checking if a key exists in metadata)

bendichter:
This looks like good changes, but I agree with @CodyCBakerPhD it would be nice to have better documentation of exactly what was fixed. A sentence or two on the motivation of each change would be very helpful. Also, it looks like PRs from external repos don't trigger tests. @luiztauffer can you run tests locally and confirm that they still pass? @alejoe91 can you adopt this PR so that we can get CI on it?

alejoe91:
@bendichter it does trigger tests on the fork: https://github.com/luiztauffer/spikeextractors/actions/runs/793286740

Passing :) not sure why they are not triggered here, I'll look into it

bendichter:
@alejoe91 great, thanks for the quick response!

luiztauffer:
The main fixes are:

**L409**
The argument for an ElectrodeGroup is `device`, not `device_name`. Metadata with device_name would fail schema validation. This same change is reflected in other changes

**L427**
`ElectrodeGroup` is an array in metadata and, if not passed by the user, the default operation should get it from the channel groups in the recording object. Some other changes reflect this change in `defaults` to a list

**L448**
`Device` is an array in metadata

bendichter:
@CodyCBakerPhD can you take a look at this?

CodyCBakerPhD:
@bendichter @luiztauffer The changes make more sense now, but the safest way to ensure that none of these break existing data interfaces would be to expand the testing infrastructure on metadata handling, both here and on the nwb-conversion-tools side. I can work on that in the coming days.

alejoe91:
@CodyCBakerPhD let me know when this is ready to merge (with added tests)

PR #628, opened 2021-04-29
issue with slicing in mearec

colehurwitz:
Sorry I have been MIA for a while (working on finishing up my thesis).  I caught this issue when helping someone getting spikeinterface working. Basically, the channel_ids are converted into a numpy array with the sort operation and will cause the slicing to fail. You have to convert it to a list to make it work. Surprisingly, casting it as a list does not work, you have to use numpy.tolist().


colehurwitz:
I don't know if this is a mearec recording extractor issue only so maybe my solution is only a bandaid and we need to fix it elsewhere? Not sure.

samuelgarcia:
Hi Cole,
normally, we don't need to propagate this to the new spikeinterface because this is done now in neo.

PR #633, opened 2021-05-10
electrodes table column add problem

Saksham20:
I'm trying to add a custom data to existing default columns in the Electrodes table of nwb and not trying to add any new column to this table. 
The metadata is like:
```python
metadata['Ecephys']['Electrodes'] = [dict(name='filtering',
                                                  description='filtering',
                                                  data=['1000Hz']*96),
                                             dict(name='group_name',
                                                  description='electrode group name',
                                                  data=self._group_name*96)
                                             ]
```
`filtering `and `group_name `are default colnames for this table, the values of which I want to override. However, the `write_electrodes()` method ends up calling `nwbfile.add_electrode_column()` for these and thus throws an error. 

This PR is for the fix that prevents this problem.


luiztauffer:
@alejoe91 you can close this one now

PR #636, opened 2021-05-11
fix ced time offset

luiztauffer:
fix for https://github.com/catalystneuro/mease-lab-to-nwb/issues/34

### DIFF
--- a/spikeextractors/extractors/cedextractors/utils.py
+++ b/spikeextractors/extractors/cedextractors/utils.py
@@ -32,6 +32,8 @@ def get_channel_info(f, smrx_ch_ind):
         Index of smrx channel. Does not match necessarily with extractor id.
     """
 
+    nMax = int(f.ChannelMaxTime(smrx_ch_ind) / f.ChannelDivide(smrx_ch_ind))
+    time_offset = f.FirstTime(chan=smrx_ch_ind, tFrom=0, tUpto=nMax) / f.ChannelDivide(smrx_ch_ind)
     ch_info = {
         'type': f.ChannelType(smrx_ch_ind),           # Get the channel kind
         'ch_number': f.PhysicalChannel(smrx_ch_ind),  # Get the physical channel number associated with this channel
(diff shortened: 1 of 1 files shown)


luiztauffer:
@alejoe91 I think this solves the ced issue

spikeextractors/extractors/cedextractors/utils.py
```diff
@@ -70,12 +73,17 @@ def get_channel_data(f, smrx_ch_ind, start_frame=0, end_frame=None):
     if end_frame is None:
         end_frame = int(f.ChannelMaxTime(smrx_ch_ind) / f.ChannelDivide(smrx_ch_ind))
 
+    nMax = int(f.ChannelMaxTime(smrx_ch_ind) / f.ChannelDivide(smrx_ch_ind))
+    frame_offset = int(f.FirstTime(chan=smrx_ch_ind, tFrom=0, tUpto=nMax) / f.ChannelDivide(smrx_ch_ind))
```
samuelgarcia, line 77: I think this is strange. For me tUpTo is in ticks so without device.

samuelgarcia:
Yep.
You are right for this line.
thanks.
I made a comment on your patch, the  `tUpto=nMax` is strange in your patch.
For me this value is in ticks no ?

spikeextractors/extractors/cedextractors/utils.py
samuelgarcia: See This https://github.com/samuelgarcia/python-neo/blob/5321483836a72e853a8f4812e796cb635cbdef77/neo/rawio/cedrawio.py#L161

luiztauffer: I agree it is strange, but this arg seems to work as a max number of frames. You can test it by fixing a small value there (e.g. 10) and you'll see that it will return max 10 frames

luiztauffer:
@samuelgarcia this documentation might be helpful in general
[SonPy.pdf](https://github.com/SpikeInterface/spikeextractors/files/6459247/SonPy.pdf)

alejoe91:
Fixed https://github.com/catalystneuro/mease-lab-to-nwb/issues/34

PR #637, opened 2021-05-13
Axonaunitextractor

sbuergers:
Depends on 

* [x] https://github.com/JuliaSprenger/python-neo/pull/1
* [x] https://github.com/JuliaSprenger/python-neo/pull/3
* [x] https://github.com/JuliaSprenger/python-neo/pull/4
* [x] https://github.com/NeuralEnsemble/python-neo/pull/986

Attempt to use Julia's neo implementation for reading `unit` data instead of using `pyxona` to generate traces from `axona` recordings. In these recordings only data around the putative spiketimes are kept. Here the interspike periods are filled Gaussian noise to create a continuous recording that can be used with `spikeinterface` to perform spike sorting. 

Tested sporadically in a notebook thus far, but needs some proper tests. Those could be in a different PR though.


samuelgarcia:
Hi Steffen,
could you also make a PR on spiekinterface for the next API ?

sbuergers:
Hi Sam,

I started doing that here https://github.com/SpikeInterface/spikeinterface/pull/152, but it is not finished. For now I have to reprioritize, finalizing the project I am working on using only the old API for the moment. I hope I can finalize it at a later stage though.

samuelgarcia:
Hi Steffen,
make the wrapper from neo should take only a few lines see this for instance:
https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/neoextractors/blackrock.py
or  this
https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/neoextractors/mearec.py
You don't have to overwrite any the method of Recording and Sorting.

alejoe91:
@samuelgarcia this is for the unit mode to make a dummy continuous signal for spike sorting!

sbuergers:
Hi @samuelgarcia , yes, for the case of `AxonaRecordingExtractor` it is not that complicated. Like Alessio said, complications mostly arise when looking at `AxonaUnitRecordingExtractor`, which has it's own special `get_traces` method for generating a dummy continuous signal. But there are also little things here and there that tripped me up when making a simple wrapper. For instance, this: https://github.com/JuliaSprenger/python-neo/pull/2#issue-647839255.

I am currently finalizing `AxonaUnitRecordingExtractor` for the old API, and I hope to get back to this once the conversion project is wrapped up. It should not be too complicated. :-)

samuelgarcia:
OK Sorry. I did not get it.

So if this code transform waveform snippets to a fake recording I would vote for a generic class not a specialized class for Axona.

In short a class that take a neo rawio that handle waveforms extraction and render on the fly trace with theses waveforms.
It will be usefull for other old dataset like spike2/plexon/.....

sbuergers:
We resolved the issue in a different way in the end, without needing to create a fake signal in AxonaRawIO. 

Since the dependencies have been merged in `neo`, should we merge this one, too? :-)

PR #644, opened 2021-05-24
added save_si function

CodyCBakerPhD:
@alejoe91 @luiztauffer Let me know if you'd like this moved to a particular corner of this ecosystem, or even a different repo.

### DIFF
--- a/spikeextractors/__init__.py
+++ b/spikeextractors/__init__.py
@@ -14,4 +14,6 @@ from .extraction_tools import load_probe_file, save_to_probe_file, read_binary, 
     write_to_h5_dataset_format, get_sub_extractors_by_property, load_extractor_from_json, load_extractor_from_dict, \
     load_extractor_from_pickle
 
+from .save_tools import save_si_object
+
 from .version import version as __version__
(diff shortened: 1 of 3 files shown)


luiztauffer:
is this only for `nwb-conversion-tools` or for a more general use?

alejoe91:
@CodyCBakerPhD I think we could leave it here! It's general quite general purpose and SI specific. Note that in the new api this dmstrategy is implemented directly in the base class

CodyCBakerPhD:
@alejoe91 Yeah, this is very particular to the legacy version of spikeextractors that will be left behind when the new API is released. Good to hear it's that ingrained in the new API though!

alejoe91:
@CodyCBakerPhD all tests are failing due to some import issues. Can you try importing the cache extractors from relative paths?

CodyCBakerPhD:
@alejoe91 OK tests should be passing now. Not actually an issue of relative imports per se but rather than having a function within the extraction_tools umbrella that imports classes like RecordingExtractor, etc, causes a circular import because those all use extraction tool functions as well. I simply moved it to a different location and it resolved the circulatirty.

PR #646, opened 2021-06-07
numpy channel ids setting

Saksham20:
functionality to manually set channel ids to a `NumpyRecordingExtractor`. I'm proposing that we use an internal attribute `self._channel_ids` that the get method returns. This way for the 1% case when the channel ids need to be changed, the internal attribute can be overridden. Its similar to what the `NeoBaseRecordingExtractor` is using. 
For a broader applicability, we could have a `set_channel_ids`() method in the `RecordingExtractor` base class. Again, this may be useful only in a minority of cases, but just an extra method and few lines of code in the base class would suffice without breaking anything downstream. What do you all think?

### DIFF
--- a/spikeextractors/extractors/numpyextractors/numpyextractors.py
+++ b/spikeextractors/extractors/numpyextractors/numpyextractors.py
@@ -41,6 +41,7 @@ class NumpyRecordingExtractor(RecordingExtractor):
 
         self._ttl_frames = None
         self._ttl_states = None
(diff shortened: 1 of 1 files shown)


alejoe91:
Hi @Saksham20 

You can already set the channel ids using the `renamed_channel_ids` arg of the `SubRecordingExtractor`, which is a more general solution applicable to all extarctors already.

PR #649, opened 2021-06-17
fix CED traces missing last element

### LINKED ISSUE #648: CEDExtractor traces are missing the last element
The trace array imported from an smrx file using CEDExtractor has one element less than when the same file is imported directly with sonpy

lkeegan:
- fix off-by-one error in num_frames, nMax, end_frame in cedrecordingextractor
- add test case for CED extractor issue: last element missing from traces
- resolves #648

### DIFF
--- a/tests/test_extractors.py
+++ b/tests/test_extractors.py
@@ -5,6 +5,7 @@ import unittest
 from pathlib import Path
 
 import numpy as np
+from sonpy import lib as sp
 
 import spikeextractors as se
 from spikeextractors.exceptions import NotDumpableExtractorError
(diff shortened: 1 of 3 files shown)


samuelgarcia:
Hi @lkeegan : could you check that the neo implementation (that will replace the spiekextractor implementation in the futur spikeinterface release) handle this correctly ?

https://github.com/NeuralEnsemble/python-neo/pull/987

Next release : https://spikeinterface.readthedocs.io/en/latest/releases/0.90.0.html

lkeegan:
@samuelgarcia the neo implementation looks good to me - I tried using `CedRawIO` from your PR with the tiny test case here and another larger smrx file & got the correct number of elements in both cases

samuelgarcia:
cool.
Thanks.
The neo will replace the spikeextractor in next release.

tests/test_extractors.py
```diff
@@ -323,6 +324,18 @@ def test_biocam_extractor(self):
...
+        trace_out = np.arange(0, n_elements, dtype=np.short)
+        assert len(trace_out) == n_elements
+        smrx_file = sp.SonFile('test.smrx')
+        smrx_file.SetWaveChannel(0, 7, sp.DataType.Adc)
+        smrx_file.WriteInts(0, trace_out, 0)
+        del smrx_file
+        recording = se.CEDRecordingExtractor('test.smrx', smrx_channel_ids=[0])
+        trace_in = recording.get_traces([0])[0]
+        assert len(trace_in) == n_elements
+
```
alejoe91, line 338: Can you make this optional on having sonpy?
```suggestion
    def test_ced_extractor(self):
        try:
            from sonpy import lib as sp
            n_elements = 881
            trace_out = np.arange(0, n_elements, dtype=np.short)
            assert len(trace_out) == n_elements
            smrx_file = sp.SonFile('test.smrx')
            smrx_file.SetWaveChannel(0, 7, sp.DataType.Adc)
            smrx_file.WriteInts(0, trace_out, 0)
            del smrx_file
            recording = se.CEDRecordingExtractor('test.smrx', smrx_channel_ids=[0])
            trace_in = recording.get_traces([0])[0]
            assert len(trace_in) == n_elements
     except:
         print("sonpy not installed")
```

samuelgarcia: sonpy is in the requirements_test.txt. So I think it is OK.
If we want to skip the test which is bad idea I think I prefer use `@pytest.skip()`

alejoe91:
Ok, merging then! Thanks @lkeegan

PR #651, opened 2021-06-23
CED set gains

### LINKED ISSUE #647: Comparison SE/neo/mat73 on spike2 files
Hey guys, 
I'm trying to work with >2GB files in Spike2, thus I'm trying to move from .smr to .smrx. 
I tried using the CED extractor here in SE, and made some comparisons with neo's CED IO recently implemented by @samuelgarcia , and a .mat version of the same file which I read with the package [mat73](https://pypi.org/project/mat73/).

My problem is, when I get the same array from these 3 different sources, the output is different and might or might not work in the rest of the analysis I have to do in the lab. (If I plot with matplotlib, the signals look the same.. except the y  range ofc)

Output: 
- from mat file using mat73: 'values': array([-1.95318462e-05, -1.35807368e-05, -1.27872554e-05, ...,
         1.35502187e-05, -1.31229592e-06,  4.66933216e-06])}

- from smrx file using SE: array([[-640., -445., -419., ...,  623.,  444.,  -43.]], dtype=float32)

- from neo: array([[-0.12800391],
       [-0.08900272],
       [-0.08380256],
       ...,
       [ 0.08880271],
       [-0.00860026],
       [ 0.03060094]], dtype=float32)

The only one that works for my further analysis (spectral analysis using [HVC](https://hybrid-vocal-classifier.readthedocs.io/en/latest/index.html)) is the first (from the mat file ... or if I export the song channel to .smr and work with neo -- that works as well)

Any comments on that?
I'm trying to write a script to get smrx data in the lab, but now I'm not sure if I should consider exporting them to .mat instead..

Figure_1

luiztauffer:
Fix #647 

- Set gain values as `gain` instead of `ced_gain`
- Set `self.has_unscaled = True` for decorator `@check_get_traces_args` to work

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: s-weigand/setup-conda@v1
         with:
-          python-version: 3.7
+          python-version: 3.8
       - name: Which python
         run: |
           conda --version
--- a/spikeextractors/cacheextractors.py
+++ b/spikeextractors/cacheextractors.py
@@ -39,6 +39,7 @@ class CacheRecordingExtractor(BinDatRecordingExtractor, RecordingExtractor):
 
         self.set_tmp_folder(tmp_folder)
         self.copy_channel_properties(recording)
+        self.copy_times(recording)
 
         if 'gain' in recording.get_shared_channel_property_names() and not return_scaled:
             self.set_channel_gains(recording.get_channel_gains())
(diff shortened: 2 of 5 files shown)


alejoe91:
Thanks Luiz! I'll test that the output is the same as NEO and fix the failing tests (due to a Sonpy import). Merging later today

alejoe91:
@luiztauffer after checking against neo it turns out that the gain is from raw to mV, so I had to ass a *1000. I also included the CED in the GIN tests

PR #656, opened 2021-07-05
return CED data as np.array instead of python list

lkeegan:
- reduces RAM usage of [get_traces](https://github.com/SpikeInterface/spikeextractors/blob/118aa9d2acb395e76bcb40c80a9db6f28a054230/spikeextractors/extractors/cedextractors/cedrecordingextractor.py#L101) (21GB -> 13GB for a small test smrx file with 32 channels)
- could be further reduced by using a smaller np data type (which it looks like [neo already does](https://github.com/samuelgarcia/python-neo/blob/master/neo/rawio/cedrawio.py#L82))

### DIFF
--- a/spikeextractors/extractors/cedextractors/utils.py
+++ b/spikeextractors/extractors/cedextractors/utils.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 try:
     from sonpy import lib as sp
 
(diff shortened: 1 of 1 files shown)


alejoe91:
Hi @lkeegan, thanks for the PR.

We are currently transitioning to the new API (spikeinterface master) which already uses NEO for most of the recording extractors. I recommend switching to the new API, as the old version (spike extractors, spiektoolkit and co.) will not be maintained after a final release in the following weeks.

lkeegan:
@alejoe91 ok will do, thanks

PR #665, opened 2021-11-24
fix to NeuroscopeSorting IDs

CodyCBakerPhD:
Ramon discovered an issue with a newly seen case of Neuroscope format; where the unit IDs in the `.txt` file are not in the usual 0-indexed order. Seeing if this fixes the issue for him.

### DIFF
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -406,6 +406,8 @@ class NeuroscopeSortingExtractor(SortingExtractor):
             )
             unit_map = dict(zip(unique_ids, list(range(n_clu))))
 
+            if 0 in unique_ids:
+               unit_map.pop(0)
             if not keep_mua_units and 1 in unique_ids:
                 unit_map.pop(1)
             self._unit_ids = unit_map.values()
--- a/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
+++ b/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
@@ -401,22 +401,17 @@ class NeuroscopeSortingExtractor(SortingExtractor):
             n_clu = clu[0]
             clu = np.delete(clu, 0)
             unique_ids = np.unique(clu)
-            if 0 not in unique_ids:  # missing unsorted IDs
-                n_clu += 1
-            if 1 not in unique_ids:  # missing mua IDs
-                n_clu += 1
+            assert len(unique_ids) == n_clu, (
+                "First value of .clu file ({clufile_path}) does not match number of unique IDs!"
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py
```diff
@@ -401,22 +401,17 @@ def __init__(
...
             clu = np.delete(clu, 0)
             unique_ids = np.unique(clu)
-            if 0 not in unique_ids:  # missing unsorted IDs
-                n_clu += 1
-            if 1 not in unique_ids:  # missing mua IDs
-                n_clu += 1
+            assert len(unique_ids) == n_clu, (
+                "First value of .clu file ({clufile_path}) does not match number of unique IDs!"
+            )
+            unit_map = dict(zip(unique_ids, list(range(n_clu))))
```
samuelgarcia, line 407: If it is possible possible to have an empty unit ( a unit with no spike) then this check could lead to problem.
In neo for instance it is possible to have unit without spikes.
In short this `unique_ids = np.unique(clu)` do not garanty to have the full unit list.

CodyCBakerPhD: This assertion is simply that the Neuroscope file is correct to its own format. The `n_clu` value is the first line of what is essentially a `.txt` file, and _should_ indicate the number of units reported in the file (including unsorted and mua). The rest of the values in the `.txt` file are the unit IDs for each spike occurrence (one for each spike occurrence, in order of time of occurrence - that info, though, is in a separate file).

For Neuroscope format it's therefore not possible to have a unit with no spikes, since as we're learning now the actual ID of each non-trivial unit is arbitrary (not incremented from 2, 3, 4, ... as previously thought). Note the trivial unit IDs of 0 (unsorted, always ignored) and 1 (mua, optional to include) are not arbitrary.

What we'd really need for both this and neo is some sample data for the format on GIN - I think that's been on Julia's to-do list for a while but not sure what the status is on it.

h-mayorquin:
I tested this today with the previously problematic session and it seems to work:
```python
    from spikeextractors import NeuroscopeMultiSortingExtractor
    path = '/shared/catalystneuro/Buzsaki/TingleyD/DT8/20170311_684um_2088um_170311_134350/'
    extractor = NeuroscopeMultiSortingExtractor(folder_path=path, keep_mua_units=False)
    extractor.get_unit_spike_train
    extractor.get_unit_spike_train(0)

CodyCBakerPhD:
@samuelgarcia Fix seems to be working for us - let me know if there's anything else I can do here; unless you wanted to wait until Julia gets those sample files on GIN so we can add tests for this?

samuelgarcia:
HI @CodyCBakerPhD this is fine for me. I let Alessio check and merge this.

PR #669, opened 2021-12-17
Adding matlab reading capabilities for new format in CellExplorerExtractor

h-mayorquin:
The purpose of this pull request is to fix/update some reading problems with the reading of matlab files in the `CellExplorerExtractor` 

1) We add capabilities to consistently read `sessionInfo` and and `spikes.Info` files independently on the matlab format (that is either `scipy.io` or `hdf5storage`).
2) Add the location of the the matlab files as an optional parameter in the extractor.
3) Add the sampling frequency as an optional parameter so a reading of `SessionInfo` is not required.

### DIFF
--- a/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
+++ b/spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
@@ -42,9 +42,10 @@ class CellExplorerSortingExtractor(SortingExtractor):
         assert (
             spikes_matfile_path.is_file()
         ), f"The spikes_matfile_path ({spikes_matfile_path}) must exist!"
-        folder_path = spikes_matfile_path.parent
-        sorting_id = spikes_matfile_path.name.split(".")[0]
+        
         if sampling_frequency is None:
+            folder_path = spikes_matfile_path.parent
+            sorting_id = spikes_matfile_path.name.split(".")[0]
             if session_info_matfile_path is None:
                 session_info_matfile_path = folder_path / f"{sorting_id}.sessionInfo.mat"
 
@@ -55,14 +56,13 @@ class CellExplorerSortingExtractor(SortingExtractor):
             try:
                 session_info_mat = scipy.io.loadmat(file_name=str(session_info_matfile_path))
                 self.read_session_info_with_scipy = True
-            except:  # <- Find out exactly what error does this produce - don't use bare excepts
+            except NotImplementedError:
                 session_info_mat = hdf5storage.loadmat(file_name=str(session_info_matfile_path))
                 self.read_session_info_with_scipy = False
-           
-           # Check the assert and extraction logic depending on read_session_info_with
+            
             assert session_info_mat["sessionInfo"]["rates"][0][0]["wideband"], (
                 "The sesssionInfo.mat file must contain "
-                "a 'sessionInfo' struct with field 'rates' containing field 'wideband'!"
+                "a 'sessionInfo' struct with field 'rates' containing field 'wideband' to extract the sampling frequency!"
             )
             if self.read_session_info_with_scipy:
                 self._sampling_frequency = float(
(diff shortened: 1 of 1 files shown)


spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -26,41 +29,78 @@ class CellExplorerSortingExtractor(SortingExtractor):
...
+        ), f"The spikes_matfile_path ({spikes_matfile_path}) must exist!"
         folder_path = spikes_matfile_path.parent
         sorting_id = spikes_matfile_path.name.split(".")[0]
+        if sampling_frequency is None:
+            if session_info_matfile_path is None:
+                session_info_matfile_path = folder_path / f"{sorting_id}.sessionInfo.mat"
+
+            assert (
+                session_info_matfile_path.is_file()
+            ), "No sessionInfo.mat file found in the folder!"
```
alejoe91, line 53:
```suggestion
            ), f"No {sorting_id}.sessionInfo.mat file found in the folder!" 
```

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -26,41 +29,78 @@ class CellExplorerSortingExtractor(SortingExtractor):
...
+            if self.read_session_info_with_scipy:
+                self._sampling_frequency = float(
+                    session_info_mat["sessionInfo"]["rates"][0][0]["wideband"][0][0][0][0]
+                )  # careful not to confuse it with the lfpsamplingrate; reported in units Hz
+            else:
+                self._sampling_frequency = float(
+                    session_info_mat["sessionInfo"]["rates"][0][0]["wideband"][0][0]
+                )  # careful not to confuse it with the lfpsamplingrate; reported in units Hz
+        else:
+            self._sampling_frequency = sampling_frequency
```
alejoe91, line 76:
```suggestion
            assert sampling_frequency is not None, "Could not load sampling frequency from files. Use the 'sampling_frequency' argument"
            self._sampling_frequency = sampling_frequency
```

alejoe91 (changes requested):
@h-mayorquin Thanks for these fixes! I had just a couple of small comments

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
```diff
@@ -26,41 +29,79 @@ class CellExplorerSortingExtractor(SortingExtractor):
...
+            if self.read_session_info_with_scipy:
+                self._sampling_frequency = float(
+                    session_info_mat["sessionInfo"]["rates"][0][0]["wideband"][0][0][0][0]
+                )  # careful not to confuse it with the lfpsamplingrate; reported in units Hz
+            else:
+                self._sampling_frequency = float(
+                    session_info_mat["sessionInfo"]["rates"][0][0]["wideband"][0][0]
+                )  # careful not to confuse it with the lfpsamplingrate; reported in units Hz
+        else:
+            assert sampling_frequency is not None, "Could not load sampling frequency from files. Use the 'sampling_frequency' argument"
```
h-mayorquin, line 76: @alejoe91  you suggested this assert but I am confused about it. The program only enters this else when the sampling_frequency is not None, so why check for it again?

h-mayorquin (commented):
I removed some comments and formatting issues (I have to say there are some aspects of black here that I don't like but I kept them that way).

h-mayorquin:
@alejoe91 what else would be required for this to be merged?

spikeextractors/extractors/cellexplorersortingextractor/cellexplorersortingextractor.py
alejoe91: Right, you can remove this else ;)

h-mayorquin:
> > @alejoe91 what else would be required for this to be merged?
> 
> Looks good! Waiting for tests and merging.

All right, removed the assert. It seems that the tests were failing tough. I will check that out.

alejoe91:
> All right, removed the assert. It seems that the tests were failing tough. I will check that out.

Don't worry about it. It seems some weird dependency issue with tridesclous. I'll fix it ;)

PR #673, opened 2022-04-18
Warning reduction : remove numpy type warnings.

h-mayorquin:
Very simpler PR to remove the deprecation warning discussed in here:

https://github.com/catalystneuro/nwb-conversion-tools/pull/459

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -3,7 +3,7 @@ dependencies:
   - python=3.8
   - pip
   - pip:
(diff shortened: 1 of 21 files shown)


h-mayorquin:
I am frankly confused on what went on here with this error. It seems unlikely that this small modification messed up the numpy dependency.

h-mayorquin:
OK, now it works. I had to include the suggestions in #672
So this should also close #672

PR #675, opened 2022-06-30
Remove comparision deprecation warning

h-mayorquin:
Remove deprecation warnings from comparison with unequal channels.

### DIFF
--- a/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
+++ b/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py
@@ -118,7 +118,7 @@ class BinDatRecordingExtractor(RecordingExtractor):
     @check_get_traces_args
     def get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_scaled=True):
         if self._complete_channels:
-            if np.all(channel_ids == self.get_channel_ids()):
+            if np.array_equal(channel_ids, self.get_channel_ids()):
                 traces = self._timeseries[:, start_frame:end_frame]
             else:
                 channel_idxs = np.array([self.get_channel_ids().index(ch) for ch in channel_ids])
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
@alejoe91 I know this is in LEGACY status, but the CI here also shows what we found on NCT with regards to the new SpikeGLX file on GIN

@samuelgarcia if you think back-compatibility of testing infrastructure is worth it, maybe add back the original `Noise4Sam` in a different directory (that we can avoid installing on the modern tests) so it can be used previously?

Or another possibility perhaps, is it possible to install particular versions (identified I assume by the hashes) of a GIN repo through datalad? @yarikoptic @JuliaSprenger

h-mayorquin:
Thanks, @CodyCBakerPhD 

Related:
https://github.com/catalystneuro/nwb-conversion-tools/pull/560

JuliaSprenger:
Hi @CodyCBakerPhD I am not sure what exactly you are aiming to fix here, but yes you can checkout a specific branch or commit accessing git via datalad, e.g. `dataset.repo.call_git(['checkout', 'main'])`, where dataset is your datalad dataset instance, as for example here: https://github.com/SpikeInterface/spikeextractors/blob/dff4dccfdd156b15d96608b0865e5895a743ea9f/tests/test_gin_repo.py#L23

yarikoptic:
> Or another possibility perhaps, is it possible to install particular versions (identified I assume by the hashes) of a GIN repo through datalad? @yarikoptic @JuliaSprenger

in case of no submodules: @JuliaSprenger 's answer is a good one! Also there is `git_clone_opts=None,` option to datalad's `clone` if that is the one you use  -- then you could pass `git_clone_opts=["-b", "main"]`.  if you have submodules -- install top dataset into correct branch and then install subdatasets.
But what problem you are trying to solve?

CodyCBakerPhD:
> But what problem you are trying to solve?

version A of the dataset contains a file structured in a certain way that the software here is built around, which is a small part of upstream tests as well.

version B of the dataset took that file and altered it in ways that break this software, so now the tests here (and upstream) fail.

We can't update the actual software here to be compatible here because it's in LEGACY status, so it makes more sense to 'freeze' the version of the testing dataset that this testing suite uses.

